Issue #106, opened 2026-07-16
[Bug]: local-cloak navigate hard-codes waitUntil 'load'; Page.goto never forwards waitUntil, so never-idle sites (zillow.com) always time out


ngaurav:
## Summary

Adapters cannot opt out of waiting for the browser `load` event during navigation. The local-cloak runtime hard-codes `waitUntil: 'load'` in its `navigate` action, and `Page.goto()` never forwards its `waitUntil` option to the daemon — it only uses it to skip the client-side settle/stealth exec. On sites that stream analytics requests indefinitely (e.g. zillow.com), the `load` event never fires and every `page.goto()` in the adapter times out after 30s, even though the document rendered long before.

## Reproduction

```bash
webcmd zillow rentals "San Francisco, CA" --max-rent 5000 --min-beds 2 --pets --parking
```

Fails with:

```
ok: false
error:
  code: UNKNOWN
  message: "page.goto: Timeout 30000ms exceeded.
    - navigating to \"https://www.zillow.com/homes/for_rent/San-Francisco,-CA_rb/\", waiting until \"load\""
```

The retained trace (traceId `20260716091350-631878c5`) shows ~200 network events completed and the final screenshot contains fully rendered search results — the page was usable; only the `load` event stalled (a TikTok analytics pixel and similar requests keep the network busy indefinitely).

Passing `{ waitUntil: 'none' }` from the adapter has no effect on the navigation itself, confirming the option is not forwarded.

## Root cause

1. `src/browser/runtime/local-cloak/actions.ts:84` — the `navigate` action hard-codes the Playwright wait condition:
   ```ts
   await lease.page.goto(command.url, { waitUntil: 'load' });
   ```
2. `src/browser/page.ts:106` — `Page.goto()` sends only `url` (plus command opts) in the `navigate` command; `options.waitUntil` is consulted afterwards only to decide whether to run the settle/stealth exec, so the daemon-side wait condition is unreachable from adapters.

## Expected behavior

`page.goto(url, { waitUntil: 'none' })` (or a `domcontentloaded`-style option) should propagate through the `navigate` command so the runtime uses a matching Playwright wait condition. Adapters that gate readiness on their own selector waits should not be held hostage by third-party analytics stalling `load`.

## Actual behavior

Navigation always waits for `load` with a 30s timeout regardless of adapter options; sites with never-idle network fail even though the DOM is ready.

## Workaround (verified)

[...]

PR #212, opened 2026-08-04
fix: honor waitUntil when opening a tab, not just when navigating

### LINKED ISSUE #210: ``waitUntil: 'load'is still hardcoded innewPage— the #106 fix reachednavigateonly, sotab new --url still hangs on never-idle sites
# Description

#106 reported `navigate` hard-coding `waitUntil: 'load'`. That fix landed, but the same root cause has a second call site that was not covered.

On `main` today:

```ts
// src/browser/runtime/local-cloak/actions.ts:116  — navigate, fixed by #106/#107
await lease.page.goto(command.url, { waitUntil: command.waitUntil === 'none' ? 'commit' : 'load' });

// src/browser/runtime/local-cloak/session-manager.ts:249  — newPage, still hardcoded
await acquired.page.goto(input.url, { waitUntil: 'load' });
```

So `webcmd browser navigate --wait-until none` now works, while `webcmd browser tab new --url <url>` (and the hosted tab-open path) still blocks until `load` fires. On a site that never goes idle — a streaming dashboard, a long-poll app shell, an ad-heavy page with a hanging subresource — opening a tab still hangs, which is the exact failure #106 was filed about.

## Why it was missed

The gap is structural rather than a missed edit. `newPage` has no `waitUntil` in its signature at all:

```ts
// src/browser/runtime/local-cloak/session-manager.ts:237
async newPage(input: SessionKeyInput & { url?: string }): Promise<CloakPageLease>
```

And its only caller drops the field, even though it is available on the command object it is reading from:

```ts
// src/browser/runtime/local-cloak/actions.ts:156-163  — 'tabs' / 'new'
const lease = await manager.newPage({
  profileId: resolveCloakCommandProfileId(manager, command),
  session: command.session,
  surface: command.surface,
  siteSession: command.siteSession,
  idleTimeout: command.idleTimeout,
  url: command.url,
  windowMode: command.windowMode,
  // command.waitUntil is never passed
});
```

Every other layer is already plumbed for this:

| Layer | Status |
| --- | --- |
| `src/browser/protocol.ts:36` | `waitUntil?: 'load' \| 'none'` already on the command type |
| `src/browser/base-page.ts:167` | abstract `goto` already accepts `waitUntil` |
| `src/browser/cdp.ts:210` | honours it (`:215`, `:221`) |
| `src/browser/page.ts:103` | honours it (`:107`, `:134`) |
| `session-manager.ts:237` | **the only type in the chain missing it** |

## Suggested fix

Three small edits, all in the local-cloak runtime:

1. Widen the `newPage` input to `SessionKeyInput & { url?: string; waitUntil?: 'load' | 'none' }` (`session-manager.ts:237`).
2. Pass `waitUntil: command.waitUntil` from the `tabs`/`new` branch (`actions.ts:156-163`).
3. Use it at `session-manager.ts:249` instead of the hardcoded literal.

## Related inconsistency worth folding in

`actions.ts:116` translates `'none'` into Playwright's `'commit'`, while `cdp.ts:215` and `page.ts:134` branch on the string `'none'` directly. Two vocabularies for one concept across three files is part of why the second call site was easy to overlook. A single shared helper — `toGotoWaitUntil(waitUntil)` — would make the next such change land everywhere at once.

## Verification

Checked against `main` (`53310bf`) and `upstream/main`; both are identical for every file named here. A full sweep of `waitUntil` across `src/` (excluding tests) shows `session-manager.ts:249` as the only remaining hardcoded `'load'` on a `goto` call.

rajarshidattapy:
## Description

`waitUntil` was plumbed through every layer of the local Cloak runtime except one: `newPage`. #106 reported `navigate` hard-coding `waitUntil: 'load'` and that fix landed in `actions.ts`, but the same root cause had a second call site that was never covered.

Before this PR:

```ts
// actions.ts — navigate, fixed by #106/#107
await lease.page.goto(command.url, { waitUntil: command.waitUntil === 'none' ? 'commit' : 'load' });

// session-manager.ts — newPage, still hardcoded
await acquired.page.goto(input.url, { waitUntil: 'load' });
```

On a site that never goes idle — a streaming dashboard, a long-poll app shell, a page with a hanging subresource — opening a page still blocked on the load event even when the caller explicitly asked to skip it.

Closes #210

### Why it survived the first fix

`newPage` had no `waitUntil` in its signature, and its only caller dropped the field even though it was reading it off the same command object:

```ts
// actions.ts — 'tabs' / 'new'
const lease = await manager.newPage({
  profileId: resolveCloakCommandProfileId(manager, command),
  session: command.session,
  surface: command.surface,
  siteSession: command.siteSession,
  idleTimeout: command.idleTimeout,
  url: command.url,
  windowMode: command.windowMode,
  // command.waitUntil was never passed
});
```

Every other layer was already plumbed — `protocol.ts` carries `waitUntil?: 'load' | 'none'` on `BrowserRuntimeCommand`, and `base-page.ts`, `cdp.ts`, and `page.ts` all accept and honor it. `newPage`'s input type was the only link in the chain missing it.

### Approach

The three-line version of this fix would copy the `'none' ? 'commit' : 'load'` ternary into `newPage`. That mapping duplicated across two files is exactly what let this bug survive the #106 fix, so instead it lives in one exported helper that both call sites use:

```ts
/**
 * Map the protocol's navigation wait condition onto Playwright's `goto` option.
 * 'none' becomes 'commit': sites that stream analytics forever never fire the
 * load event, so callers gating readiness on their own selector waits must be
 * able to skip it. Every `goto` in this runtime routes through here so a new
 * call site cannot quietly reintroduce a hardcoded 'load'.
 */
export function toGotoWaitUntil(waitUntil?: 'load' | 'none'): 'load' | 'commit'
```

## Scope change after merging main

This PR was opened against a `main` that still exposed `webcmd browser tab new --url <url>` as a CLI command. That surface no longer exists: the public browser surface is now `tabs`, `bind`, `run`, `snapshot`, and `close`, and `tabs` is list-only.

The CLI-facing half of the original PR has therefore been dropped as obsolete:

- the `--wait-until` option on `tab new` and its `resolveBrowserWaitUntil` helper in `cli.ts`
- the matching `cli.test.ts` cases
- the `tab/new` option contract in `command-catalog.ts`
- the `docs/cli-reference.mdx` "Browser Runtime" section and the `SKILL.md` `browser tab new` row

The runtime bug itself is unaffected by that restructure and is still live on `main` today (`session-manager.ts` still hardcodes `waitUntil: 'load'`, `actions.ts` still drops `command.waitUntil`). The path is now reached through the adapter-facing `IPage.newTab()` API rather than from the CLI, so the fix still matters — it just no longer needs a CLI flag attached to it.

### Changes

| File | Change |
| --- | --- |
| `session-manager.ts` | Added `toGotoWaitUntil()`; widened `newPage` input with `waitUntil?: 'load' \| 'none'`; replaced the hardcoded `'load'` |
| `actions.ts` | `tabs`/`new` now forwards `command.waitUntil`; `navigate` switched to the shared helper |
| `page.ts`, `types.ts` | `newTab(url?, options?: { waitUntil?: 'load' \| 'none' })` |
| `page.test.ts`, `provider.test.ts` | Tests for the forwarding and the commit-only wait |

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Hosted mode

This fix is local-Cloak only. Hosted `newTab` ignores the wait condition entirely, and hosted `navigate` maps `'none'` to `domcontentloaded` rather than `commit`, so hosted behavior still diverges after this merges. Tracked separately in agentrhq/webcmd-cloud#30 — this PR should not be read as closing that gap.

## Verification

Run against `main` merged into this branch:

```
tsc --noEmit                clean
npm test                    4532 passed | 1 skipped (378 files)
npm run build               OK
check:hosted-contract       PASS
check:typed-error-lint      PASS
check:silent-column-drop    PASS
check:package-bin           PASS
```

`check:hosted-contract` is unchanged because `command-catalog.ts` now matches `main` exactly.

No behavior change for the default path: omitting `waitUntil`, or passing `'load'`, still waits for the load event exactly as before. Only the explicit `'none'` case differs.

The `Cannot find module 'impit'` error mentioned in the original description was a stale local `node_modules` and does not reproduce on a clean install. No action needed.


ankitranjan7:
The leftover `waitUntil` hole on tab-open landed in #316. This PR overlaps that fix — worth closing unless it still has unique coverage in `page.ts`/`types.ts`.

ankitranjan7:
Thank you @rajarshidattapy — `waitUntil` on Cloak `tab new --url` already landed on `main` as #316 (`fix(local-cloak): honour waitUntil when opening a tab with a url`).

Closing as superseded; your diagnosis was right, and the later PR kept the same fix against current `main`.

ankitranjan7:
Superseded by #316.

PR #227, opened 2026-08-07
fix(browser): allow fixture to override maxColumns and maxNestedDepth

### LINKED ISSUE #218: webcmd browser verify: validateRowShape() hardcodes a 12-column cap with no override, so it can never pass for wide-row adapters
## What's wrong

`webcmd-adapter-author`'s SKILL.md (Step 12 / Runbook) marks a passing `webcmd browser verify` fixture as **required** for every adapter, with no documented exception:

> `verify/<cmd>.json`: **required.** Expected values for `webcmd browser verify`: args, rowCount, columns, types, patterns, notEmpty.

But `validateRowShape()` in `@agentrhq/webcmd/dist/src/browser/verify-fixture.js:205` hardcodes:

```js
const DEFAULT_MAX_TOP_LEVEL_KEYS = 12; // verify-fixture.js:33
const maxTopLevelKeys = opts.maxTopLevelKeys ?? DEFAULT_MAX_TOP_LEVEL_KEYS;
```

and the call site in `cli.js:2715` invokes it with no options at all:

```js
const shapeFailures = validateRowShape(rows);
```

There is no CLI flag to raise `maxTopLevelKeys`. Any adapter whose rows have more than 12 top-level keys **cannot ever pass** `webcmd browser verify`, regardless of how correct the adapter is.

## Impact

This isn't a hypothetical edge case — it affects a whole adapter family already shipped in the marketplace. The university postgraduate-course-export adapters (`cincinnati`, `yale`, `concordia`, `goettingen`, `heidelberg`, `hft`, `iit`) all emit a 52-column row shape (by design — the output is meant to become a spreadsheet/CSV for a downstream team, not a narrow agent-queryable API). None of them ship a verify fixture, presumably because it's structurally impossible to satisfy.

A subagent building a new adapter in this family (Manchester) followed the skill literally, hit repeated verify failures, tried `--write-fixture`, and eventually had to read webcmd's own dist source to figure out why — before falling back to `webcmd validate`, which is the gate that actually applies to `browser:false` plugin-style adapters.

## Suggested fix

One of:
- Add a `--max-top-level-keys` (or similar) override flag to `webcmd browser verify`, and have the skill tell agents when a wide row shape is legitimate (e.g. spreadsheet-export adapters) vs. a smell.
- Have the skill explicitly document the exception: for `browser:false`/export-style adapters with an intentionally wide row shape, `webcmd validate <site>` is the required gate, not `webcmd browser verify`.

## Context

Found while a subagent built a new university-course-export adapter (Manchester) following `webcmd-adapter-author` end to end. Happy to share the full transcript excerpt if useful.

SparshM8:
## Description

Resolves #218. 

This PR updates the `FixtureExpect` schema and the `validateRowShape` logic to respect `maxColumns` and `maxNestedDepth` overrides. Previously, the browser verify step strictly hardcoded the caps to 12 columns and a nesting depth of 1, which caused legitimate spreadsheet-export adapters to fail CI. 

By adding these options, adapter authors can now explicitly permit wider rows or deeper nesting in their `verify/<cmd>.json` expectations without disabling shape validation entirely.

Related issue: #218

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [x] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

### Adapter Notes

## Screenshots / Output

N/A - CLI output remains the same, but adapters with wide rows will now pass `webcmd browser verify` without throwing shape errors.


ankitranjan7:
Thanks for working on this! The fixture-level override was a useful idea, but we’re merging the smaller CLI flag fix in #230. Closing this in favor of #230 🙌

PR #230, opened 2026-08-07
fix: allow overriding the verify row-shape top-level key cap

### LINKED ISSUE #218: webcmd browser verify: validateRowShape() hardcodes a 12-column cap with no override, so it can never pass for wide-row adapters

ayushsingh82:
## Summary
- `validateRowShape()` (`src/browser/verify-fixture.ts`) already accepted an optional `maxTopLevelKeys`, but `webcmd browser verify` never exposed a way to set it — it always called `validateRowShape(rows)` with the hardcoded 12-key default. Any adapter whose rows are wide by design (the shipped 52-column university course-export adapters, for example) could never pass verify regardless of correctness.
- Adds a `--max-top-level-keys <n>` flag to `browser verify`, validated as a positive integer, threaded through to `validateRowShape(rows, { maxTopLevelKeys })`. Default behavior (12) is unchanged when the flag is omitted.
- Updates the failure message to reflect the actual configured cap and to point authors at the new flag.
- Updates the adapter-author skill's fixture checklist (Step 12) so agents building wide-row adapters know to use `--max-top-level-keys` instead of skipping the required verify fixture.

Fixes #218

## Test plan
- [x] Unit: `validateRowShape` respects a raised `maxTopLevelKeys` and still rejects keys beyond it (`src/browser/verify-fixture.test.ts`)
- [x] Integration: `browser verify` rejects a 20-key row by default, passes with `--max-top-level-keys 20`, and rejects a non-positive value with a usage error (`src/cli.test.ts`)
- [x] `npm test` — 385 test files / 4625 passed, 1 skipped (unchanged skip, +3 new passing tests)

## Note for maintainers
While verifying test coverage for this fix I noticed `vitest.config.ts`'s `unit` project excludes `src/browser/**/*.test.ts` entirely (added in the #216 plugin-migration merge), and no other configured project or CI job covers that glob either — so all 44 test files under `src/browser/` (including `verify-fixture.test.ts`, `cdp.test.ts`, `analyze.test.ts`, etc.) currently don't run in CI. That's unrelated to this fix, so I'm not touching `vitest.config.ts` here, but wanted to flag it — happy to file a separate issue if useful.


ayushsingh82:
@ngaurav Fixed — this PR adds a `--max-top-level-keys <n>` override to `webcmd browser verify` so wide-row adapters (like the 52-column university course-export adapters) can pass verify, closing #218.

ankitranjan7:
Thanks for the PR!

PR #232, opened 2026-08-08
fix: enforce plugins/<site> registration in root webcmd-plugin.json

### LINKED ISSUE #222: webcmd-adapter-author skill: root webcmd-plugin.json step is documented but skippable — should be enforced, not optional
## Correction

I originally filed this suggesting the root `webcmd-plugin.json` registration step might be unnecessary and could be removed from the skill. That was the wrong read — flagging the correction: the fix should go the other way. The step is meant to be mandatory whenever a plugin is promoted into the repo; the actual bug is that the skill lets it be silently skipped with no error and no warning.

## What the skill says

`references/adapter-template.md` in the `webcmd-adapter-author` skill documents this promotion sequence for turning a private CLI into a repo plugin (also referenced from `SKILL.md` line 260):

```bash
webcmd plugin create <site> --dir plugins/<site> --description "<site> commands for Webcmd"
cp ~/.webcmd/clis/<site>/*.js plugins/<site>/
rm plugins/<site>/hello.ts plugins/<site>/greet.ts 2>/dev/null || true
```

Then add `<site>` to the root `webcmd-plugin.json` `plugins` map:

```json
"<site>": {
  "path": "plugins/<site>",
  "version": "0.1.0",
  "description": "<site> commands for Webcmd",
  "webcmd": ">=0.2.0"
}
```

## What actually happens

Across three independent adapter builds in the same repo (manchester, warwick, mdxdubai), the root-manifest step was skipped every single time. Nothing in `webcmd plugin create`, `webcmd plugin install file://...`, or `webcmd validate <site>` errors, warns, or otherwise signals that a required step was missed — the commands all succeed and the plugin works, so there's no feedback loop that would catch the omission. The repo now has three `plugins/<site>/webcmd-plugin.json` files (one per plugin) but no root-level `webcmd-plugin.json` registering any of them.

## Why this matters

The skill presents the root-manifest edit as a required part of the promotion sequence, but nothing enforces it — an agent (or a human following the same doc) can complete the entire promotion flow, see every command succeed, and never know the repo-level registration was left out. That's a silent, discoverable-only-by-omission gap, which is worse than either (a) the tooling itself failing loudly if the manifest is missing, or (b) the skill treating it as a real checklist item with a verification step.

## Suggested fix

- Add an explicit verification step to `adapter-template.md` right after the root-manifest edit — e.g. `grep -q '"<site>"' webcmd-plugin.json || echo "MISSING: <site> not registered in root webcmd-plugin.json"` — so an agent following the skill has to actively confirm the step landed, not just perform an edit and move on.
- Consider whether `webcmd validate <site>` or `webcmd plugin install` could themselves warn when a plugin exists under `plugins/<site>` but isn't listed in a root `webcmd-plugin.json`, so the gap is caught by tooling instead of relying on an agent following documentation to the letter.

ayushsingh82:
## Summary
- Promoting a CLI to a repo plugin has an easy-to-skip step: after `webcmd plugin create` scaffolds `plugins/<site>/`, the author must also register `<site>` in the root `webcmd-plugin.json` `plugins` map. Nothing in `webcmd plugin create`, `webcmd plugin install file://...`, or `webcmd validate <site>` errors or warns when that registration is skipped — the plugin works locally either way. Per the issue, three separate adapter builds (manchester, warwick, mdxdubai) all missed it silently.
- Adds `scripts/check-plugin-root-registration.mjs`: fails when a `plugins/<site>` directory has no matching root-manifest entry, and also catches the reverse (a root-manifest entry with no matching directory). Mirrors the style/location of the existing `scripts/check-plugin-command-parity.mjs`.
- Adds `check:plugin-root-registration` to `package.json`.
- Adds a verification step to the `adapter-template.md` promotion sequence so an agent following the skill catches the gap immediately, not just at CI time.

Fixes #222

## Not included: CI wiring
I'd normally add this as a step in `.github/workflows/ci.yml` right next to the existing "Check plugin command parity" step, but my fork's token doesn't have the `workflow` scope GitHub requires to push workflow-file changes. A maintainer can add:

```yaml
      - name: Check plugin root registration
        if: runner.os == 'Linux'
        run: npm run check:plugin-root-registration
```

right after the `check:plugin-parity` step in the `build` job (`.github/workflows/ci.yml`) — happy to open a follow-up once I sort out the token scope, whichever is easier for you.

## Test plan
- [x] Verified the check fails when a `plugins/<site>` dir exists without a root-manifest entry (simulated with a throwaway fixture dir, removed after)
- [x] Verified the check fails the other direction too (a root-manifest entry with no matching dir)
- [x] Verified it passes clean against the current repo (123 plugins, all registered)
- [x] `npm test` — 385 test files / 4623 passed, 1 skipped (unchanged baseline)

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
     "check:package-bin": "node scripts/check-package-bin.mjs",
     "check:silent-column-drop": "node scripts/check-silent-column-drop.mjs",
     "check:typed-error-lint": "node scripts/check-typed-error-lint.mjs",
-    "check:plugin-parity": "node scripts/check-plugin-command-parity.mjs"
+    "check:plugin-parity": "node scripts/check-plugin-command-parity.mjs",
+    "check:plugin-root-registration": "node scripts/check-plugin-root-registration.mjs"
   },
   "keywords": [
     "cli",
--- a/scripts/check-plugin-root-registration.mjs
+++ b/scripts/check-plugin-root-registration.mjs
@@ -1,0 +1,59 @@
+#!/usr/bin/env node
+/**
+ * check-plugin-root-registration.mjs — every plugins/<site>/ must be listed
+ * in the root webcmd-plugin.json plugins map, and vice versa.
+ *
+ * The promotion flow (webcmd-adapter-author skill / adapter-template.md) has
+ * an easy-to-skip step: after `webcmd plugin create` scaffolds plugins/<site>/,
+ * the author must also add <site> to the root webcmd-plugin.json. Nothing in
+ * `webcmd plugin create`, `webcmd plugin install file://...`, or
+ * `webcmd validate <site>` errors or warns when that step is skipped — the
+ * plugin works locally either way, so the omission is silent until someone
+ * tries `webcmd plugin install github:agentrhq/webcmd/<site>` (which
+ * resolves sub-plugins by looking up <site> in the root manifest) and it
+ * fails to resolve. This check catches the gap at CI time instead. See #222.
+ */
+
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const root = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const pluginsDir = path.join(root, 'plugins');
+const rootManifest = readJson(path.join(root, 'webcmd-plugin.json'));
+const registered = new Set(Object.keys(rootManifest.plugins ?? {}));
+
+const onDisk = fs
+  .readdirSync(pluginsDir, { withFileTypes: true })
+  .filter((entry) => entry.isDirectory())
+  .map((entry) => entry.name)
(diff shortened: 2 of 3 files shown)


ayushsingh82:
@ngaurav Fixed — this PR adds `scripts/check-plugin-root-registration.mjs` (catches a plugins/<site> directory with no matching root webcmd-plugin.json entry, or vice versa) plus a verification step in the adapter-template.md promotion sequence, closing #222. CI wiring is noted in the PR description since my fork's token currently lacks the `workflow` scope to push that part myself.

ankitranjan7:
Thanks for working on this. I’m closing it because the repository workflow has changed: root `webcmd-plugin.json` registration and the README community table are generated from `plugins/*/webcmd-plugin.json` after merge by the community-plugin sync workflow. Contributors should not edit or separately enforce those generated files. #117 will update the adapter-author guidance accordingly.

PR #234, opened 2026-08-08
docs: make CLI-promotion confirmation an explicit, hard-to-read-past gate

### LINKED ISSUE #223: webcmd-adapter-author skill promotes CLIs to repo plugins without an explicit user confirmation gate

ayushsingh82:
## Summary
- The adapter-author skill gated promoting a private CLI into a repo plugin behind "When the user says to promote a CLI" — a soft clause inside a longer sentence. Per the issue, across three separate adapter builds a general instruction like "build a real, working adapter, this will be run by the user once done" was read as implicit authorization for the entire pipeline (private iteration + promotion + install), with no separate checkpoint asking whether to promote.
- Promotion writes into the shared repo and is the moment an adapter goes from drafted to shipped, so it needs its own explicit ask rather than being satisfied implicitly by "build a working adapter."
- Rewrites the `SKILL.md` "Key Conventions" bullet as a standalone, bolded gate: private iteration alone fully satisfies "build a working adapter," `webcmd plugin create`/promotion must wait for explicit confirmation, and the agent should state the checkpoint out loud before promoting rather than continuing silently.
- Mirrors the same note at the top of `references/adapter-template.md`'s promotion section, since that's the other place an agent following the skill lands on the promotion commands.

Fixes #223

## Test plan
- [x] Doc-only change; no tests reference these skill files (`grep -rl` came up empty)
- [x] `npm test` — 385 test files / 4623 passed, 1 skipped (unchanged baseline)


ayushsingh82:
@ngaurav Fixed — this PR turns the soft "when the user says to promote a CLI" clause into an explicit, bolded confirmation gate in SKILL.md's Key Conventions (and mirrors it in adapter-template.md), so a general "build a working adapter" instruction can no longer be read as implicit authorization to promote, closing #223.

Issue #231, opened 2026-08-07
vitest unit project excludes src/browser/**/*.test.ts entirely — 44 test files never run in CI


ayushsingh82:
## What's wrong

`vitest.config.ts`'s `unit` project excludes the entire `src/browser/**/*.test.ts` glob:

```ts
{
  test: {
    name: 'unit',
    include: ['src/**/*.test.ts'],
    exclude: ['src/browser/**/*.test.ts'],
    sequence: { groupOrder: 0 },
  },
},
```

This was added in the #216 plugin-migration merge (commit `0470ac43`). No other configured project picks up the slack:

- `plugin` only includes `plugins/*/test/**/*.test.{ts,js}`
- `e2e-fixed-port` / `e2e` only include specific named files under `tests/e2e/`
- `smoke` only includes `tests/smoke/**/*.test.ts`

I checked every job in `.github/workflows/*.yml` that runs vitest (`unit`, `plugin`, `smoke`, `e2e`, `e2e-fixed-port`, plus the bun variant) — none of them cover `src/browser/**`.

## Impact

`find src/browser -name "*.test.ts" | wc -l` → **44 test files**, including `verify-fixture.test.ts`, `cdp.test.ts`, `analyze.test.ts`, `snapshot/*.test.ts`, `run/*.test.ts`, `runtime/local-cloak/*.test.ts`, etc. None of these run under `npm test`, `npm run test:all` (still project-scoped via the same config), or any CI job. Regressions in this whole subsystem currently won't be caught automatically.

Confirmed locally: running any of these files under an ad-hoc vitest config (bypassing the project filters) passes fine — so they're not failing, they're just not being invoked.

## Suggested fix

Either narrow the `unit` project's exclude to whatever specifically motivated it during the #216 migration (if only some browser tests needed temporary exclusion), or add an explicit project/include that covers `src/browser/**/*.test.ts` (mirroring `unit`'s config, or its own project if these need different setup).

## Context

Found while verifying test coverage for #230 (`--max-top-level-keys` fix), which added tests to `src/browser/verify-fixture.test.ts` — those tests are correct (verified via a standalone ad-hoc vitest config) but currently don't execute in CI due to this exclusion.

ankitranjan7:
Verified fixed on main — closing.

PR #238, opened 2026-08-08
fix: run src/browser unit tests in CI and stabilize the vendor digest

Agnik47:
## Description

The `unit` vitest project excluded `src/browser/**/*.test.ts`, and no other project picked those files up — `plugin` covers `plugins/*/test/**`, `e2e` lists individual files under `tests/e2e/`, and `smoke` covers `tests/smoke/**`. The result was that **44 test files / ~596 tests** covering CDP, snapshots, the Playwright sandbox and the local Cloak runtime ran in no CI job at all.

Removing the exclude (now: collapsing the `include` list back to `src/**/*.test.ts`) also restores vitest's default excludes (`node_modules`, `dist`) for that project, and matches the other four projects, none of which set `exclude`.

One small correction to the issue: the exclude came from #196, not #216. #216 only dropped the `clis/**` half of the list during the adapter migration, and `b841842` later rewrote the exclusion as an include-glob without closing the gap:

| commit | `unit` project shape |
| --- | --- |
| `c08cb276` (init) | `include: ['src/**/*.test.ts']`, `exclude: ['clis/**/*.test.{ts,js}']` |
| `b51cac37` (#196) | `exclude: ['clis/**/*.test.{ts,js}', 'src/browser/**/*.test.ts']` |
| `0470ac43` (#216) | `exclude: ['src/browser/**/*.test.ts']` |
| `b841842` | `include: ['src/*.test.ts', 'src/!(browser)/**/*.test.ts', 'src/browser/verify-fixture.test.ts']` (exclude dropped, same gap re-expressed as an include glob) |
| this PR | `include: ['src/**/*.test.ts']` |

### Why the two extra changes are here

Enabling the tests surfaces a pre-existing failure in `src/browser/run/playwright-client-build.test.ts`, which shells out to `build-playwright-sandbox-client.mjs --check`. Without fixing it, this PR would turn the `windows-latest` unit job red. Two independent causes:

**1. The vendor digest was platform-dependent.** `vendorDigest()` hashes each vendored file's relative path, built with `path.join()`, so entries hash as `client\artifact.ts` on Windows and `client/artifact.ts` everywhere else. The check could therefore never pass on Windows:

```
as-is (native separator) : 882f0880d4838423647cc617ff199298536b606803fd8dbcdb3cb5a8bed31131
POSIX-normalized         : 23a97afd31faa0e5a64b116e897c3b90123eafcc2ca9102586a062eaf533350d
vendor-manifest.json     : 23a97afd31faa0e5a64b116e897c3b90123eafcc2ca9102586a062eaf533350d
```

Normalizing the separator to `/` is a no-op on POSIX, so **the pinned hash is unchanged** and `vendor-manifest.json` needs no regeneration.

**2. The vendored sources and generated bundle are compared byte-for-byte.** On a checkout with `core.autocrlf=true` (the Git for Windows default), CRLF breaks both the digest and the separate "bundle is up to date" comparison — the latter with a misleading `Generated Playwright QuickJS client is out of date` error. Upstream `main` has since gained its own repo-wide `.gitattributes` (`* text eol=lf`), which already covers this, so my scoped version isn't needed anymore and this PR no longer touches `.gitattributes`.

Both fixes are necessary; I verified each one independently fails without the other.

Related issue: #231

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [x] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

`npm run typecheck`, `npm run build` and the `git diff --exit-code` generated-artifact check all pass.

All 44 previously-skipped files now run and pass through the real `unit` project (Windows, Node 22):

```
$ npx vitest run --project unit src/browser/
 Test Files  44 passed (44)
      Tests  598 passed | 2 skipped (600)
```

Full `unit` project, before and after — same pre-existing failures, no new ones. I confirmed these 6 files fail identically on **unmodified upstream/main** in a side worktree, so none are caused by this PR:

```
before (unmodified main): Test Files  6 failed | 96 passed (102)   Tests  18 failed | ~1614 passed
after (this branch):      Test Files  6 failed | 139 passed (145)  Tests  18 failed | 2212 passed | 7 skipped (2237)
```

The 6 failing files (`discovery`, `docs-sync-review-cli`, `plugin`, `hosted/runner`, `hosted/files`, `site-memory/local-store`) are local-machine artifacts, not real failures: the symlink-related ones need Windows Developer Mode/admin (`EPERM: operation not permitted, symlink`) to create test fixtures, and `hosted/runner`/`hosted/files`/`docs-sync-review-cli` fail because my temp dir lives under `$HOME`, so paths render as `~\AppData\...`. Both reproduce on clean `main` and pass in CI.

### One thing I could not verify

I don't have Bun locally, so I could not run the `bun-test` job, which also uses `--project unit` and will now execute these 44 files. `QuickJSHost` (WASM via `quickjs-emscripten`) and the test that spawns `process.execPath` look like the likeliest friction points under Bun. Flagging it rather than assuming — happy to adjust if that job goes red.

### Notes

- No new test added deliberately: the guard for the digest regression is `playwright-client-build.test.ts` itself, which now actually runs on Windows in CI. Adding a dedicated test would mean exporting `vendorDigest()` out of the provenance script purely for testability — happy to do it if you'd prefer.
- Rebased onto current `main` to pick up `b841842`'s include-based rewrite of the `unit` project and the new repo-wide `.gitattributes`; re-verified typecheck, build, the generated-artifact diff, and both the scoped `src/browser` and full `unit` project vitest runs after rebasing. Diff is now 2 files instead of 3.
- Existing Windows clones will see the vendored files re-checkout as LF once this lands (already true today via upstream's `.gitattributes`, independent of this PR).


ankitranjan7:
Thanks again for this contribution, @Agnik47. I updated the branch with the current `main` and kept both parts of the PR intact.

The unit-test discovery now excludes only `src/browser/runtime/local-cloak/browser-run.test.ts`, which is a Chromium-backed integration suite and cannot run in the unit CI jobs because they do not install a browser. All other `src/**/*.test.ts` files remain automatically discovered.

Local verification on Node 22 is green: clean install, build, 440 test files (5,473 passed / 56 skipped), and the Playwright sandbox digest check. The complete GitHub matrix is also green, including Bun and the macOS, Ubuntu, and Windows build, unit, and plugin jobs.

Issue #242, opened 2026-08-08
[Bug]: closing one background profile kills profiles whose directory name shares a prefix


Kaushik2003:
### Description

On macOS, background profile teardown selects which Chromium processes to kill
with a substring test on the `ps` command line, so it also matches unrelated
profiles.

`terminateProfile` — `src/browser/runtime/local-cloak/darwin-background-launch.ts:42-50`:

```ts
const needle = `--user-data-dir=${userDataDir}`;
for (const line of stdout.split('\n')) {
  if (!line.includes(needle)) continue;
  const pid = Number.parseInt(line.trim().split(/\s+/, 1)[0], 10);
  if (Number.isInteger(pid) && pid !== process.pid) process.kill(pid, 'SIGTERM');
}
```

Profiles are siblings under `~/.webcmd/cloak/profiles/<id>` (`profiles.ts:23`) and
ids are validated only by character class, `/^[A-Za-z0-9._-]+$/`, so hyphens and
dots are allowed. `--user-data-dir=…/profiles/work` is therefore a substring of
`--user-data-dir=…/profiles/work-2`, and tearing down `work` sends `SIGTERM` to the
browser serving `work-2`. Prefix-sharing names are ordinary: `default` /
`default-2`, `actor` / `actor-profile`, `test` / `test.old`.

There is also no check that the matched process belongs to webcmd, so a user's own
Chrome started with a matching `--user-data-dir` is killed too.

Two things make this easy to hit rather than exotic:

- `background` is the default window mode (`execution.ts:536-540`), so on macOS
  every browser-backed command goes through this launcher unless `--window
  foreground` is passed.
- `terminateProfile` runs both on normal context close
  (`darwin-background-launch.ts:112`) and on the launch-failure cleanup path
  (`darwin-background-launch.ts:118`), so one profile failing to start is enough to
  kill a healthy one.

The victim gets no useful signal — its next command fails with the generic
`Target page, context or browser has been closed`, which is likely why this has
gone unnoticed.

`session-manager.ts:603-668` already solves exactly this problem for lock recovery:
`commandUsesProfileDir` requires the directory to end at an argument boundary,
`profileDirAliases` resolves symlinks, and `isCloakBrowserCommand` restricts the
match to CloakBrowser binaries. `terminateProfile` uses none of it.

### Steps to Reproduce

1. On macOS in local mode, start background browsers under two profiles whose
   names share a prefix:
   ```bash
   webcmd --profile work github whoami
[...]

Issue #225, opened 2026-08-06
Profile launch issue


rishabhraj36:
# Background-mode browser launch fails, and concurrent profiles surface a cryptic error (CloakBrowser free-tier session cap)

## Summary

Two related problems make multi-profile / adapter-eval runs unusable, both surfacing as the same cryptic error:

```
page.goto: Target page, context or browser has been closed
```

After tracing it end to end, the webcmd daemon and profile multiplexing are **correct** — the failures originate in the CloakBrowser launch layer:

1. **Background mode (default) self-terminates on launch.** A single background CloakBrowser instance exits ~250–320ms after the daemon connects, even with only one profile and nothing else running. Every command that then does a `page.goto` fails with `Target page, context or browser has been closed`.
2. **Concurrent profiles hit CloakBrowser's free-tier 1-session cap.** Launching a second profile's browser while a first is alive makes the second exit with code **76**. This is a CloakBrowser licensing limit, not a webcmd bug — but webcmd reports it as the same opaque "target closed" error with no hint that it's a concurrency/licensing issue.

The net effect: the twitter adapter eval (`webcmd-adapter-eval`, actor/peer profiles) fails on every step, and the error message points nowhere useful.

## Environment

- webcmd: `v0.5.2`
- cloakbrowser: `0.4.5`
- Chromium binary: `chromium-150.0.7871.114.3-pro`
- OS: macOS (Apple Silicon)
- CloakBrowser license tier: Free (no `CLOAKBROWSER_LICENSE_KEY` / `license.key`)

### Case 1 — background mode, single profile

```
webcmd daemon restart
webcmd --profile actor-profile twitter timeline --limit 5
```

Result:

```
ok: false
error:
  code: COMMAND_EXEC
  message: "Pre-navigation to https://x.com failed: page.goto: Target page, context or browser has been closed ..."
```

### Case 2 — two profiles concurrently (foreground)

```
export WEBCMD_WINDOW=foreground
webcmd --profile peer-profile twitter whoami     # keeps peer's browser alive
webcmd --profile actor-profile twitter login     # second live browser
```

The second launch exits 76 and `page.goto` fails.

## Evidence (daemon instrumentation)

With logging added at the launch and context-`close` points, plus `DEBUG=pw:browser`:

**Background, single profile — browser disconnects on its own ~320ms after connect:**

```
[cloak-debug] bg connected port=59315 dir=.../peer-profile           at 03:40:08.380Z
[...]

PR #243, opened 2026-08-08
fix(local-cloak): match profile processes on an exact --user-data-dir boundary

Kaushik2003:
## Description

On macOS, `terminateProfile` picked the Chromium processes to `SIGTERM` with a bare
substring test on the `ps` command line
(`darwin-background-launch.ts:42-50`). Profiles are siblings under
`~/.webcmd/cloak/profiles/<id>` and ids may contain `-` and `.`, so
`--user-data-dir=…/profiles/work` matched `…/profiles/work-2` as well: tearing down
one background profile killed a healthy concurrent one. With no CloakBrowser binary
check, a user's own Chrome on a matching `--user-data-dir` was killed too. The
victim only ever surfaced `Target page, context or browser has been closed`.

Lock recovery in `session-manager.ts` already matched profile processes correctly —
requiring the flag value to end at an argument boundary, resolving realpath aliases,
and restricting matches to CloakBrowser binaries. This moves that matching into a
new `profile-processes.ts` and routes both call sites through it, so there is one
implementation instead of two that disagree.

The helpers cannot live in either existing module: `session-manager.ts` already
imports `darwin-background-launch.ts`, so either direction would create a cycle.
The new module is a straight move — no behavior change on the recovery path.

Teardown on the context-close path is now best-effort like the launch-failure path
beside it, so a `ps` or `kill` failure no longer replaces the outcome of
`browser.close()`.

Files in the commit:

- `src/browser/runtime/local-cloak/profile-processes.ts` — new; profile process
  discovery and signalling, moved from `session-manager.ts`.
- `src/browser/runtime/local-cloak/session-manager.ts` — imports the moved helpers.
- `src/browser/runtime/local-cloak/darwin-background-launch.ts` — `terminateProfile`
  uses the shared matching; teardown made best-effort.
- `src/browser/runtime/local-cloak/profile-processes.test.ts` — new; regression
  coverage.

Related issue: https://github.com/agentrhq/webcmd/issues/242

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

- [x] I ran the checks relevant to this PR — `npm run typecheck`, plus the affected
      `src/browser/runtime/local-cloak` suites (see below)
- [x] I updated tests or docs if needed — added `profile-processes.test.ts`; no
      user-facing surface changed, so no docs or skills update

### Adapter Notes

Not an adapter change — this touches the local CloakBrowser runtime only.

## Screenshots / Output

The two boundary tests fail against the previous matching and pass with this
change, so they are genuine regression coverage.

```text
$ npx vitest run src/browser/runtime/local-cloak/profile-processes.test.ts --reporter=verbose

 ✓ commandUsesProfileDir > does not match a profile dir that is a prefix of another
 ✓ commandUsesProfileDir > matches the flag at the end of the command line
 ✓ isCloakBrowserCommand > ignores a Chromium that is not the Cloak build
 ✓ findCloakProfileProcesses > returns only the Cloak process owning the exact profile dir
 ✓ findCloakProfileProcesses > reports no processes when ps fails

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Reverting `commandUsesProfileDir` to the old substring test reproduces the bug in
those tests:

```text
FAIL  commandUsesProfileDir > does not match a profile dir that is a prefix of another
      AssertionError: expected true to be false

FAIL  findCloakProfileProcesses > returns only the Cloak process owning the exact profile dir
      AssertionError: expected [ 32818, 32819, 32820 ] to deeply equal [ 32818 ]
                               (sibling profile and the user's own Chrome also matched)
```

The neighbouring suites still pass, confirming the move did not change recovery
behaviour:

```text
$ npx vitest run src/browser/runtime/local-cloak/{profile-processes,session-manager,darwin-background-launch}.test.ts

 Test Files  3 passed (3)
      Tests  38 passed (38)

$ npm run typecheck
(clean)
```

**Note on CI:** these tests live under `src/browser/**`, which the vitest `unit`
project excludes — see #231 and PR #238. Until that lands they need a local vitest
config to run; CI will not report them.

**Note on #225:** distinct root cause (that one is in the CloakBrowser launch
layer), but it produces the same opaque
`Target page, context or browser has been closed` in concurrent-profile workflows,
so it may be a second contributing cause of what #225 describes.

### DIFF
--- a/src/browser/runtime/local-cloak/darwin-background-launch.ts
+++ b/src/browser/runtime/local-cloak/darwin-background-launch.ts
@@ -7,6 +7,7 @@ import { buildLaunchOptions, humanizeBrowser } from 'cloakbrowser';
 import type { LaunchPersistentContextOptions } from 'cloakbrowser';
 import { chromium } from 'playwright-core';
 import type { Browser, BrowserContext } from 'playwright-core';
+import { findCloakProfileProcesses, signalPids } from './profile-processes.js';
 
 const execFileAsync = promisify(execFile);
 
@@ -40,13 +41,7 @@ export async function waitForDevToolsPort(portFile: string, timeoutMs = 10_000):
 }
 
 async function terminateProfile(userDataDir: string): Promise<void> {
-  const { stdout } = await execFileAsync('/bin/ps', ['-axo', 'pid=,command=']);
-  const needle = `--user-data-dir=${userDataDir}`;
-  for (const line of stdout.split('\n')) {
-    if (!line.includes(needle)) continue;
-    const pid = Number.parseInt(line.trim().split(/\s+/, 1)[0], 10);
-    if (Number.isInteger(pid) && pid !== process.pid) process.kill(pid, 'SIGTERM');
-  }
+  signalPids(await findCloakProfileProcesses(userDataDir), 'SIGTERM');
 }
 
 const defaultDependencies: Dependencies = {
@@ -109,7 +104,9 @@ export async function launchDarwinBackgroundPersistentContext(
         await browser!.close();
       } finally {
         contextActivators.delete(context);
-        await deps.terminateProfile(options.userDataDir);
+        // Best-effort: a ps/kill failure must not replace the outcome of
+        // `browser.close()`, which is what the caller awaited.
+        await deps.terminateProfile(options.userDataDir).catch(() => {});
       }
     };
     return context;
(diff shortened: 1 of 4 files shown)


Kaushik2003:
@ngaurav - Tagging you for visibility. Happy to adjust this fix if it doesn't align with the vision for the project.

adikulkarni006 (approved):

ankitranjan7:
Superseded by #291 and shipped in WebCMD 0.6.2. Current `process-matcher.ts` provides stricter exact profile and Cloak executable matching. The best-effort teardown catch remains potentially useful, but should be submitted as a focused follow-up.

Issue #247, opened 2026-08-10
[Bug]: web fetch-browser is not installed by default, so the FETCH_BLOCKED escalation path is a dead end


ankitranjan7:
### What happens

`webcmd web fetch` is built into the CLI (client-owned path, `src/main.ts:74`). `webcmd web fetch-browser` is not — it lives in the `web` plugin, which a fresh `npm install -g @agentrhq/webcmd` does not include.

```
$ webcmd web fetch-browser --url https://example.com
Site "web" is not installed.
Search: webcmd plugin search web
```

### Why it matters

The blocked path points straight at the missing command. `src/fetch/client.ts` raises `FETCH_BLOCKED` / `FETCH_REQUIRES_BROWSER` with `hint: 'Use webcmd web fetch-browser for this URL.'`, and `skills/smart-search/SKILL.md` names it as the mandatory second rung of the cost order. On a default install the sequence is:

1. `web fetch` → `FETCH_BLOCKED`, hint says use `fetch-browser`
2. `web fetch-browser` → `Site "web" is not installed`
3. agent has no escalation path and falls back to looping search engines

Observed in a real session: a research run hit DuckDuckGo (empty), Bing (CAPTCHA), Reddit plain HTTP (WAF block), and had no working browser escalation, because the command the errors kept recommending was not installed.

### Options

- Ship `web fetch-browser` in the core CLI alongside `web fetch` — they read as one command pair, and the split is invisible until it fails.
- Or, if it must stay a plugin: make the `FETCH_BLOCKED` hint detect that `web` is not installed and emit the install command instead of naming an unavailable one, and have `smart-search` preflight availability before promising the escalation.

The asymmetry itself is the surprise — `webcmd web fetch` working while `webcmd web fetch-browser` reports the `web` site missing is hard to predict from the outside.

beubax:
Closed by #295. The old `web fetch-browser` adapter path was removed entirely: `web fetch` is now an inbuilt/client-owned command, and browser rendering is an explicit Session + `browser run` workflow rather than a missing default adapter.

Issue #246, opened 2026-08-10
[Bug]: webcmd web fetch prints a raw Node stack trace instead of a formatted CLI error


ankitranjan7:
### What happens

`webcmd web fetch` is dispatched on the client-owned fast path in `src/main.ts:74`, which calls `runClientOwnedWebFetch` directly. That function (`src/fetch/command.ts:36`) has no try/catch, and the call site is outside the `CliError` handling every other command routes through. Any thrown `CliError` therefore escapes to Node's default handler.

Reproduced on current `main` (0.5.4 build):

```
$ webcmd web fetch --url https://example.com
file:///…/dist/src/fetch/client.js:67
                throw new CliError('FETCH_BLOCKED', 'The site blocked non-browser fetches.', …);
                      ^
CliError: The site blocked non-browser fetches.
    at webFetch (file:///…/dist/src/fetch/client.js:67:23)
    at async runClientOwnedWebFetch (file:///…/dist/src/fetch/command.js:40:20)
    at async file:///…/dist/src/main.js:74:9 {
  code: 'FETCH_BLOCKED',
  hint: 'Use webcmd web fetch-browser for this URL.',
  exitCode: 1
}
```

### Expected

The same shape every other command produces, so the `code` and `hint` are the signal and the internals are not:

```
Error: The site blocked non-browser fetches. (FETCH_BLOCKED)
Hint: Use webcmd web fetch-browser for this URL.
```

### Why it matters

`web fetch` is the first call `smart-search` makes, so this is often the first error an agent sees in a session. A stack trace with absolute file paths and `at async` frames is many tokens of noise, buries the `hint` that tells the agent what to do next, and invites the agent to treat a routine block as a crash in webcmd itself.

### Fix

Wrap `runClientOwnedWebFetch` in the same error handling the rest of the CLI uses, or move the client-owned path inside the existing handler. Applies to `ArgumentError` from `clientOptions()` too.

beubax:
Closed by #295. The fix moved `web fetch` through the core command path so argument and runtime failures now use the standard structured CLI error envelope instead of leaking raw Node stack traces.

PR #263, opened 2026-08-10
fix: make client-owned web fetch discoverable (#252)

### LINKED ISSUE #252: webcmd web fetch is invisible to --help, list, completions, and -h despite always being available
### What happens

`webcmd web fetch` always works — it is intercepted on the client-owned fast path in `src/main.ts:79` before adapter discovery and before the hosted-mode branch, so it runs on a fresh install with zero plugins and in hosted mode. But nothing in the CLI's own surface says it exists.

On current `main` (0.5.4 build), with the `web` plugin not installed:

```
$ webcmd --help            # no `web` under Commands or Site adapters
$ webcmd list              # no web fetch
$ webcmd list -f json      # 0 entries with "site": "web"
$ webcmd --get-completions webcmd web
                           # (empty)
$ webcmd web fetch --help
ArgumentError: --url must be an http or https URL
    at clientOptions (…/dist/src/fetch/command.js:34:15)
```

`hosted-contract.json` has `web/fetch-browser` but no `web/fetch` entry either.

Meanwhile the command is fully functional:

```
# Example Domain
…
```

### Why this happens

There are two registration paths for the same command and only one of them feeds discovery:

- `src/main.ts:79-82` — the fast path. Hardcoded `argv[0] === 'web' && argv[1] === 'fetch'`, hand-rolled argv parsing in `clientOptions()`. Always available. Never touches commander, so it contributes nothing to help, `list`, `cli-manifest.json`, or completions, and has no `-h`.
- `plugins/web/fetch.js` — a one-line side-effect import (`import '@agentrhq/webcmd/fetch/command'`) that registers `webFetchCommand` from `src/fetch/command.ts:5` in the registry. This is what makes the command *visible* — but only for users who have installed the `web` plugin.

So availability and discoverability are decoupled: the command works for everyone, and is listed only for the subset who installed a plugin they did not need in order to use it.

### Why it matters

`web fetch` is the first call `skills/smart-search/SKILL.md` mandates, and the skill ships in the npm package (`files` includes `skills/**`). An agent that follows the documented practice of reading live command metadata before executing — `webcmd list -f json`, `webcmd <site> <command> -h`, exactly what smart-search instructs elsewhere — will conclude the command does not exist, or will try `-h` and get a stack trace telling it `--url` is required.

The `--help` output also advertises `Run 'webcmd list' for full command details`, which is inaccurate for this command.

Related but distinct: #247 is the mirror-image asymmetry (`fetch-browser` is *listed* in the hosted contract but *not installed* by default). #246 covers the stack-trace formatting on the fast path — this issue is that `-h` is not a supported argument at all, not that its failure prints badly.

### Suggested fix

Register the `web` site in the built-in program so `web fetch` appears in `--help`, `list`, `cli-manifest.json`, and completions without a plugin install, and so `webcmd web fetch -h` prints real help.

Keep the `src/main.ts` fast path as-is. The comment above it is explicit that hosted mode must not read `~/.webcmd/clis` or local site memory just to decide what commands exist, and the fast path is what guarantees `web fetch` stays client-owned in hosted mode. The two paths should stay separate; only the discovery side needs adding.

Two follow-ons if this lands:

- Decide whether `web/fetch` belongs in `hosted-contract.json`. It is client-owned, so it may warrant an explicit availability mode rather than a plain entry — `npm run check:hosted-contract` will need to agree either way.
- Once help is real, `clientOptions()` and the `cli({...})` arg spec in `src/fetch/command.ts` are two hand-maintained copies of the same flags (`--url --timeout --max-chars --allow-private`). Worth collapsing so help cannot drift from what the fast path actually parses.

ankitranjan7:
Fixes #252.

## The problem

`webcmd web fetch` always works, but the CLI never tells you it exists.

The command is intercepted on the client-owned fast path in `src/main.ts:79`, before adapter discovery — so it runs on a fresh install with zero plugins. It just isn't listed anywhere, and asking for help throws:

```
$ webcmd --help                       # no `web` under Commands or Site adapters
$ webcmd list                         # no web fetch
$ webcmd list -f json                 # 0 entries with "site": "web"
$ webcmd --get-completions webcmd web # (empty)
$ webcmd web fetch --help
ArgumentError: --url must be an http or https URL
```

So the only way to discover a built-in command is to already know it.

## What I changed

1. `clis/web/fetch.js` — registers the command through a new `makeWebFetchCommand()` factory, so both `build-manifest` and the runtime filesystem scan pick it up.
2. `src/discovery.ts` — the module pattern now matches `make<Pascal>Command(`, the same convention `src/build-manifest.ts:48` already used.
3. `cli-manifest.json` — gains the generated `web/fetch` entry. It regenerates byte-identical, `check:hosted-contract` passes, and `hosted-contract.json` is unchanged because this command is not browser-based.
4. `src/fetch/command.ts` — now that help is real, the fast path honours what it advertises: `-f`/`--format` for output (markdown still the default), structured `--help -f yaml|json`, an error on an unsupported format instead of silently falling back to a table, and an error when `--timeout`/`--max-chars` is given a flag-shaped value instead of quietly coercing it to `1`.

The fast path itself stays exactly where it was — execution never touches the registry, so hosted mode still never cloud-routes this command.

## Proof

| Surface | Before | After |
|---|---|---|
| `webcmd --help` | no `web` | `web` under Site adapters |
| `webcmd list` | absent | `fetch [public] … [builtin]` |
| `webcmd list -f json` | absent | full entry with all four args |
| `webcmd web fetch -h` | `ArgumentError` | real help |
| `webcmd --get-completions web ''` | empty | `fetch`, `fetch-browser` |

Verified against a rebuilt `dist`. Four new tests in `src/fetch/command.test.ts` cover the `-f` output path, structured help, the format rejection and the `--timeout -5` rejection. Full suite passes except `tests/e2e/plugin-management.test.ts`, which clones plugins over the network and fails identically on `main` (fixed separately in #267).

## Not in this PR

- #246 — `ArgumentError` still prints a raw stack trace on the fast path.
- #247 — `clis/` missing from `package.json` `files`. Unreachable here while the `argv[0] === 'web' && argv[1] === 'fetch'` prefix holds, and pre-existing for `fetch-browser`. Fixed in #271.

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -1,6 +1,47 @@
 [
   {
     "site": "web",
+    "name": "fetch",
+    "description": "Fetch a URL locally without launching a browser",
+    "access": "read",
+    "strategy": "public",
+    "browser": false,
+    "args": [
+      {
+        "name": "url",
+        "type": "string",
+        "required": true,
+        "help": "http(s) URL to fetch"
+      },
+      {
+        "name": "timeout",
+        "type": "int",
+        "default": 30,
+        "required": false,
+        "help": "Fetch budget in seconds"
+      },
+      {
+        "name": "max-chars",
+        "type": "int",
+        "default": 50000,
+        "required": false,
+        "help": "Maximum characters of extracted content"
+      },
+      {
+        "name": "allow-private",
+        "type": "boolean",
+        "default": false,
+        "required": false,
+        "help": "Allow private/loopback addresses"
+      }
+    ],
+    "defaultFormat": "md",
+    "type": "js",
+    "modulePath": "web/fetch.js",
+    "sourceFile": "web/fetch.js"
+  },
+  {
+    "site": "web",
     "name": "fetch-browser",
     "description": "Fetch any web page and export as Markdown",
     "access": "read",
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -36,6 +36,7 @@
       }
     ],
     "defaultFormat": "md",
+    "clientOwned": true,
     "type": "js",
     "modulePath": "web/fetch.js",
     "sourceFile": "web/fetch.js"
--- a/clis/web/fetch.js
+++ b/clis/web/fetch.js
@@ -1,1 +1,11 @@
-import '@agentrhq/webcmd/fetch/command';
+/**
+ * Discovery entry for client-owned `web fetch`.
+ *
+ * Execution stays on the main.ts fast path so hosted mode never cloud-routes
+ * this command. This module exists so build-manifest, `webcmd list`,
+ * completions, and Commander help can see the same registration as the
+ * always-available fast path.
+ */
+import { makeWebFetchCommand } from '@agentrhq/webcmd/fetch/command';
+
+export const command = makeWebFetchCommand();
--- a/clis/web/test/fetch-surface.test.ts
+++ b/clis/web/test/fetch-surface.test.ts
@@ -1,5 +1,9 @@
 /**
- * E2E coverage for the `web fetch` command surface (#252).
+ * Cross-surface coverage for the `web fetch` command surface (#252).
+ *
+ * Lives in the `plugin` project rather than `e2e`: it drives the built binary
+ * (so it needs dist/), and that project is the one CI runs in full on every
+ * OS. The `e2e` project runs only a named subset of browser files.
  *
  * The bug was cross-surface: the command always executed, but help, list and
  * completions did not know it existed. Unit tests on the parser cannot catch
@@ -15,9 +19,9 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 import * as os from 'node:os';
 import { fileURLToPath } from 'node:url';
-import { runCli, parseJsonOutput } from './helpers.js';
+import { runCli, parseJsonOutput } from '../../../tests/e2e/helpers.js';
 
-const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
+const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../..');
 const TEST_HOME = fs.mkdtempSync(path.join(os.tmpdir(), 'webcmd-web-fetch-e2e-'));
 const LOCAL_ENV = { HOME: TEST_HOME, USERPROFILE: TEST_HOME };
 // Hosted mode is selected by config.json, so pointing the config dir at a
--- a/src/build-manifest.ts
+++ b/src/build-manifest.ts
@@ -127,6 +127,7 @@ function toManifestEntry(cmd: CliCommand, modulePath: string, sourceFile?: strin
     ...(cmd.tags?.length ? { tags: [...cmd.tags] } : {}),
     ...(cmd.keywords?.length ? { keywords: [...cmd.keywords] } : {}),
     defaultFormat: cmd.defaultFormat,
+    ...(cmd.clientOwned ? { clientOwned: true } : {}),
     type: 'js',
     modulePath,
     sourceFile,
--- a/src/discovery.ts
+++ b/src/discovery.ts
@@ -38,8 +38,8 @@ export const USER_WEBCMD_DIR = getUserWebcmdDir();
 export const USER_CLIS_DIR = getUserClisDir();
 /** Plugins directory: ~/.webcmd/plugins/ */
 export const PLUGINS_DIR = getPluginsDir();
-/** Matches files that register commands via cli() or lifecycle hooks */
-const PLUGIN_MODULE_PATTERN = /\b(?:cli|registerSiteAuthCommands|onStartup|onBeforeExecute|onAfterExecute)\s*\(/;
+/** Matches files that register commands via cli() / factories or lifecycle hooks */
+const PLUGIN_MODULE_PATTERN = /\b(?:cli|registerSiteAuthCommands|onStartup|onBeforeExecute|onAfterExecute)\s*\(|\bmake[A-Z]\w*Command\s*\(/;
 
 function parseStrategy(rawStrategy: string | undefined, fallback: Strategy = Strategy.COOKIE): Strategy {
   if (!rawStrategy) return fallback;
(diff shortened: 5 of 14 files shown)


beubax:

Issue #252 is valid. `webcmd web fetch` is always available through the client-owned fast path, so hiding it from `--help`, `list`, completions, and structured help is an inconsistent command surface. Keeping execution on a fast path while deriving presentation metadata from the same argument definition is a reasonable direction.

Before this lands, the local/hosted ownership contract needs to be explicit.

### Blocking Cloud contract ambiguity

This PR adds `web/fetch` to `cli-manifest.json` as a normal `PUBLIC`, non-browser command. During package build, the hosted contract is generated from that manifest. Under the existing availability rules, a `PUBLIC` command without a local-only/app domain is hosted-capable.

At the same time, `main.ts` intercepts every `web fetch` invocation before `shouldUseHostedMode()`. Thus the normal CLI continues to execute the command locally even when the user selected hosted mode.

Today Cloud sees no root `clis/` directory in the 0.6.0 package, so this conflict is masked. PR #271 proposes staging the built-ins under `dist/src/clis`; once Cloud is updated to read that staged manifest, it can advertise and load `web/fetch` as a server-side default while the thin CLI still bypasses Cloud for the same grammar.

That creates two different ownership paths:

- CLI invocation: always local/client-owned;
- hosted manifest/direct API: potentially server-executed.

This contradicts the stated client-owned behavior in this PR and makes network-origin, timeout, billing, proxy, and SSRF semantics depend on how the command was reached.

### Required product decision

Please choose and encode one of these contracts:

1. **Client-owned command:** introduce/derive metadata that excludes `web/fetch` from the hosted adapter manifest/contract while still including it in local help, list, and completions; or
2. **Hosted in hosted mode:** move the hosted-mode boundary ahead of this fast path so hosted users call Cloud, and implement/test the Cloud execution contract deliberately.

Do not rely on the current missing-`clis` packaging bug to keep the command out of Cloud.

The north-star currently emphasizes a thin hosted client and server-side hosted execution, while this PR emphasizes an intentional client-owned exception. Either can work, but the exception must be represented in metadata rather than inferred from `main.ts` control flow.

### Required tests

Please add integration-level assertions for:

- top-level help lists `web`;
- `webcmd list -f json` contains exactly one `web/fetch` entry;
- completions contain `fetch`;
- `web fetch --help` and structured help do not require `--url`;
- local execution still uses the intended fast path;
- hosted-mode execution either remains provably client-owned and absent from the hosted manifest, or routes to Cloud—according to the decision above;
- the package built with #271 remains executable from its installed layout.

The unit tests added here cover the parser/presentation helpers, but the bug concerns cross-surface discoverability and ownership, so at least one end-to-end surface test is warranted.

### Scope note

This PR also expands output-format and validation behavior beyond the discoverability bug. Those changes are defensible if retained, but they should be tested as compatibility changes rather than assumed to follow automatically from Commander help. In particular, keep the hand-parsed fast path aligned with accepted common-option spellings.

Please coordinate this with #271 and the Cloud loader change rather than releasing the manifest changes independently.

ankitranjan7:
Agreed the ownership had to be encoded rather than inferred. Pushed `6790000`, taking **option 1: `web fetch` is a client-owned command.**

Reasoning, so you can push back if the north-star says otherwise: the command exists precisely because it needs no browser and no server — a thin local fast path is the feature. Routing it to Cloud (option 2) would add a network hop and a billing/proxy/SSRF surface to the one command that deliberately has none. If you'd rather it become server-owned in hosted mode, say so and I'll redo it as option 2 instead; the metadata flag is the only thing that would change.

### The contract, in metadata

- `BaseCliCommand.clientOwned?: boolean` — declared on the command, carried through `cli()` into `cli-manifest.json`;
- `deriveHostedAvailability()` returns `{ mode: 'local-only', reason: 'client-owned' }` for it, ahead of the strategy/domain rules, so `PUBLIC` + non-browser no longer implies hosted-capable;
- the generated `hosted-contract.json` now carries `availability: { mode: 'local-only', reason: 'client-owned' }` and `sessionPolicy: 'local-only'` for `web/fetch`. `hosted/runner.ts` already refuses local-only commands, so once Cloud reads the staged manifest from #271 it cannot advertise or execute a second copy.

Nothing about the decision depends on the missing-`clis` packaging bug any more — remove that bug (which #271 does) and the contract still says local-only.

### Tests

New e2e file `tests/e2e/web-fetch-discoverability.test.ts` (registered in the `e2e` project), driving the built binary:

- top-level `--help` lists `web`;
- `list -f json` contains **exactly one** `web/fetch` entry;
- `--get-completions --cursor 2 web` offers `fetch`;
- `web fetch --help` exits 0 and does not demand `--url`;
- structured help (`--help -f json`) likewise;
- `web fetch` with no `--url` fails on the **local** parser in both local and hosted mode — the hosted fixture points at an unresolvable API base URL, so a Cloud round-trip would show up as a DNS error rather than the `--url` message;
- `hosted-contract.json` marks `web/fetch` local-only/client-owned.

Plus a unit assertion in `availability.test.ts` that `clientOwned` flips availability while a plain `PUBLIC` command stays hosted.

`tsc --noEmit` clean, `check:hosted-contract` passes (generated bytes match committed), 4788 unit+plugin tests pass, 7/7 in the new e2e file.

### Scope note

Fair point on the output-format/validation work — the hand-parsed fast path now rejects an unknown `-f/--format` with the same list the help block advertises, and accepts the same spellings Commander does (`-f json`, `--format=json`, `-fjson`); that's covered in `src/fetch/command.test.ts`. The installed-layout executability assertion you asked for lives in #271's `check:package-bin`, which now loads every manifest module from the npm-installed prefix. Happy to sequence this behind #271 for release.

ankitranjan7:
Follow-up `dfd970e`: moved the new surface tests from `tests/e2e/` to `clis/web/test/fetch-surface.test.ts` (the `plugin` project). CI's `e2e` project only runs a named subset of browser files, so as written the assertions would never have run there; the `plugin` job runs in full on Linux/macOS/Windows after `npm run build`, which is what a test driving the built binary needs. Same 7 assertions, all passing.

beubax:
Thanks for working through the `web fetch` discoverability problem here — this was a real issue and the direction of making the client-owned command visible was right. We ended up landing the broader architectural fix in #295: `web fetch` is now a core/client-owned command published through the command surface/manifest, rather than being represented as an adapter under `clis/web`. That covers the help/list/completion discoverability issue from #252 while avoiding the old adapter lifecycle split. Closing this PR as superseded by #295.

Issue #264, opened 2026-08-10
[Bug]: web fetch challenge detection false-positives on CSP headers (news.ycombinator.com fails with FETCH_BLOCKED)


ankitranjan7:
### What happens

`isChallengeResponse` in `src/fetch/classify.ts:3` builds its evidence string from **all response headers plus the body**, then matches `challengeMarkers` (`cloudflare|cf-chl|datadome|perimeterx|px-captcha|akamai|captcha|just a moment|verify you are human`).

Hacker News serves a perfectly good `200` whose `content-security-policy` header mentions `https://cdnjs.cloudflare.com/` and `https://www.google.com/recaptcha/`. Both markers hit, so a healthy page is classified as a challenge.

```
plain 200 challenge=true jsShell=false
content-security-policy: default-src 'self'; script-src 'self' 'unsafe-inline' https://www.google.com/recaptcha/ … https://cdnjs.cloudflare.com/; …
```

Consequence: the plain tier already had the full 34 KB of usable HTML, but the impit ladder runs anyway (chrome, then firefox), both fetches return the *same* 200 that is again "a challenge", and the command ends in:

```
$ webcmd web fetch --url https://news.ycombinator.com
CliError: The site blocked non-browser fetches.
  hint: Use webcmd web fetch-browser for this URL.
```

The page was never blocked.

### Why it matters

Any site whose CSP allow-list names a CDN or reCAPTCHA — a very large share of the web — pays two wasted upstream fetches and then fails with a misleading `FETCH_BLOCKED` telling the agent to escalate to a browser it does not need.

### Suggested fix

Narrow the evidence. Header matching should look at the headers that actually signal a challenge (`server`, `cf-mitigated`, `cf-chl-*`, `x-datadome`, `set-cookie` names like `__cf_bm`/`datadome`) rather than every header value, and CSP/`report-to`/`link` should be excluded outright — they are allow-lists of third parties, not evidence about this response. Body matching can stay as-is.

Found while fixing #249; the two are independent (that one is about the retry ladder blowing the `--timeout` budget, this one is about the ladder running at all).

beubax:
Closed by #295. Challenge detection now ignores unrelated allow-list headers such as CSP/report-to/link and only treats relevant challenge evidence as blocking.

PR #265, opened 2026-08-10
fix: honour the web fetch timeout budget end to end (#249)

### LINKED ISSUE #249: [Bug]: web fetch --timeout is not honored across the retry ladder (10s budget took 68s)
### What happens

`--timeout` is documented as the fetch budget (`{ name: 'timeout', type: 'int', default: 30 }` in `src/fetch/command.ts`), but it appears to apply per attempt rather than to the call as a whole. `src/fetch/client.ts` retries across tiers — plain, then impit with a chrome profile, then firefox — so a slow or hanging host multiplies the wait.

Measured on current `main`:

```
$ time webcmd web fetch --url https://news.ycombinator.com --timeout 10
… 1:08.65 total
```

68 seconds against a 10 second budget. An unflagged run against the same host exceeded 120s.

### Why it matters

Agents use `--timeout` to bound a step. When it silently runs 7x over, a research loop that budgets a few blocked fetches stalls instead, and in a harness with its own tool timeout the call is killed with no structured error to act on.

### Expected

`--timeout` bounds the whole command. Either divide the budget across the retry tiers or apply it as an overall deadline that aborts the ladder, and surface a `FETCH_TIMEOUT`-style structured error when it is hit.

ankitranjan7:
Fixes #249.

## The problem

`--timeout` is meant to bound the whole `web fetch` command. It didn't: a 10-second budget took 68 seconds and then crashed the process.

```
$ time webcmd web fetch --url https://news.ycombinator.com --timeout 10
… 1:08.65 total
```

## Root cause

Not the retry ladder — that already honoured the deadline. Instrumenting `webFetch` shows all the fetching finished in 3.2s:

```
0.9s plain 200            challenge=true
0.9s impit start chrome   remaining=4088ms
2.1s impit done  chrome   200
2.1s impit start firefox  remaining=2932ms
3.2s impit done  firefox  200
3.2s closing proxy
...
67.9s UNCAUGHT This socket has been ended by the other party
```

The other 65 seconds were spent closing the proxy. `proxy.close()` waits for every connection to drain, and impit leaves connections open, so the cleanup in `finally` ignored the budget we had just enforced. Those same leftover connections later fired an unhandled `'error'` event, which killed the process instead of returning a usable error.

## What I changed

1. `src/fetch/safe-proxy.ts` — keep a list of every socket the proxy opens (including the upstream halves the HTTP server never sees) and destroy them all in `close()`. Closing now returns immediately.
2. `src/fetch/safe-proxy.ts` — socket `'error'` handlers destroy the socket instead of re-throwing, so a dead peer can no longer crash the process.
3. `src/fetch/client.ts` — an aborted fetch (`AbortError` / `TimeoutError`) is now converted into our structured `TimeoutError`, so hitting the budget gives `code: TIMEOUT` and the `TEMPFAIL` exit code instead of a raw `DOMException`.

## Proof

| | before | after |
|---|---|---|
| `web fetch --url https://news.ycombinator.com --timeout 5` | 68.5s, process crash | **3.4s**, clean error |
| host that accepts and never responds, `--timeout 3` | ~68s | **3.3s**, `TimeoutError: web fetch timed out after 3s` |

New test `src/fetch/safe-proxy.test.ts` opens a tunnel to a server that never replies and asserts `close()` finishes in under a second — it hangs until the test timeout without this change. Two more in `client.test.ts`: the proxy is closed when the ladder throws, and an aborted fetch surfaces as `TIMEOUT`.

Full suite passes except `tests/e2e/plugin-management.test.ts`, which fails identically on `main` (fixed separately in #267).

## Not in this PR

- Hacker News still runs the ladder pointlessly and ends in `FETCH_BLOCKED` — that is a separate defect (#264, fixed in #266). This PR bounds the time regardless of how many tiers run.
- Reading the response body has no timeout of its own (impit's `timeout` covers the request, not the body stream). Left a `ponytail:` comment naming the limit rather than building a race for a case we have not seen.

### DIFF
--- a/src/fetch/client.test.ts
+++ b/src/fetch/client.test.ts
@@ -36,4 +36,21 @@ describe('webFetch', () => {
       createSafeProxy: async () => ({ url: 'http://proxy', close: async () => {} }),
     })).rejects.toMatchObject({ code: 'TIMEOUT', message: 'web fetch timed out after 5s' });
   });
+  it('reports an impit-shaped deadline as a structured timeout', async () => {
+    // impit reports its own deadline as a plain Error — no TimeoutError/AbortError
+    // name to match on — so the budget having elapsed is what identifies it.
+    const impitTimeout = new Error('error sending request for url (https://example.com/): operation timed out');
+    await expect(webFetch({ url: 'https://example.com', timeoutSeconds: 0.05, maxChars: 0, allowPrivate: false }, {
+      plainFetch: vi.fn().mockImplementation(async () => { await new Promise(done => setTimeout(done, 80)); throw impitTimeout; }),
+      createImpit: vi.fn(),
+      createSafeProxy: async () => ({ url: 'http://proxy', close: async () => {} }),
+    })).rejects.toMatchObject({ code: 'TIMEOUT', message: 'web fetch timed out after 0.05s' });
+  });
+  it('does not relabel a failure that happened with budget left', async () => {
+    await expect(webFetch({ url: 'https://example.com', timeoutSeconds: 30, maxChars: 0, allowPrivate: false }, {
+      plainFetch: vi.fn().mockRejectedValue(Object.assign(new Error('connect ECONNREFUSED'), { code: 'ECONNREFUSED' })),
+      createImpit: vi.fn(),
+      createSafeProxy: async () => ({ url: 'http://proxy', close: async () => {} }),
+    })).rejects.toThrow('connect ECONNREFUSED');
+  });
 });
--- a/src/fetch/client.test.ts
+++ b/src/fetch/client.test.ts
@@ -19,4 +19,21 @@ describe('webFetch', () => {
     expect(createImpit).toHaveBeenNthCalledWith(1, expect.objectContaining({ browser: 'chrome' }));
     expect(createImpit).toHaveBeenNthCalledWith(2, expect.objectContaining({ browser: 'firefox' }));
   });
+  it('closes the safe proxy even when the ladder throws', async () => {
+    const close = vi.fn().mockResolvedValue(undefined);
+    await expect(webFetch({ url: 'https://example.com', timeoutSeconds: 5, maxChars: 0, allowPrivate: false }, {
+      plainFetch: vi.fn().mockRejectedValue(new Error('boom')),
+      createImpit: vi.fn(),
+      createSafeProxy: async () => ({ url: 'http://proxy', close }),
+    })).rejects.toThrow('boom');
+    expect(close).toHaveBeenCalledOnce();
+  });
+  it('reports an aborted fetch as a structured timeout', async () => {
+    const abort = Object.assign(new Error('The operation was aborted'), { name: 'TimeoutError' });
+    await expect(webFetch({ url: 'https://example.com', timeoutSeconds: 5, maxChars: 0, allowPrivate: false }, {
+      plainFetch: vi.fn().mockRejectedValue(abort),
+      createImpit: vi.fn(),
+      createSafeProxy: async () => ({ url: 'http://proxy', close: async () => {} }),
+    })).rejects.toMatchObject({ code: 'TIMEOUT', message: 'web fetch timed out after 5s' });
+  });
 });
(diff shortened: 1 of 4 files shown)


adikulkarni006 (approved):

beubax:

Issue #249 is valid, and the diagnosis is substantially correct. The retry ladder already shares a deadline, but `server.close()` can wait for Impit's keep-alive CONNECT tunnels to drain. Tracking both client and upstream sockets and destroying them during teardown is the correct layer to repair that hang.

Two lifetime/error gaps remain.

### 1. A DNS completion can create a socket after cleanup

Both the HTTP and CONNECT paths await DNS before creating their upstream connection. `close()` currently:

1. destroys the sockets present in the set;
2. clears the set;
3. calls `server.close()`.

There is no closed/closing state. If DNS resolution is still pending when cleanup starts, the handler can resume afterward and create/track a new upstream socket. Because the set was already drained, that late socket is not destroyed by the close pass and may again keep the process alive or issue an outbound connection after the fetch budget has expired.

Please add a closing barrier:

- mark the proxy closing before destroying sockets;
- have `track()` immediately destroy any socket presented after that point;
- check the state after awaited DNS and before opening/using an upstream connection;
- make repeated `close()` calls safe if that is part of the returned interface contract.

The smallest regression test is an injected lookup that does not resolve until after `proxy.close()` begins. Assert that no late tunnel survives and close completes within the bound.

### 2. The timeout normalization test models the wrong error shape

`asFetchError()` maps only errors whose `name` is `TimeoutError` or `AbortError`. The added unit test creates exactly that mocked shape, so it proves the branch but not the pinned Impit behavior.

With the pinned Impit version, a deadline can surface as a generic `Error` whose message contains its timeout condition rather than one of those names. That path still leaks a generic error instead of the structured Webcmd `TIMEOUT` envelope.

Please add a representative Impit-shaped timeout test (or a focused integration diagnostic using the pinned client), then normalize that known shape without turning unrelated network failures into timeouts. An elapsed-deadline check at the catch boundary may be safer than broad message matching, provided it cannot mislabel an early failure.

### Existing tests

The idle CONNECT test is useful and should stay: it proves that already-established tunnels no longer make `close()` wait indefinitely. It does not cover the pending-DNS race.

The “closes the safe proxy when the ladder throws” test verifies the existing `finally` behavior, not the new socket lifecycle. Please keep it if desired, but it cannot substitute for the delayed-lookup regression.

### Scope and security

This remains SSRF-sensitive code. The proposed socket destruction does not weaken address validation, but the late-resolution path must not be allowed to open a connection after teardown. Preserve DNS validation and all existing private-address checks.

Today this affects client-owned `web fetch`, so users need a new CLI release; a Cloud deploy is not required for the current dispatch path.

Once the closing-state race and real timeout normalization are covered, the implementation direction is sound.

ankitranjan7:
Both gaps addressed in `b8132d0`.

### 1. Closing barrier for the late-DNS race

`createSafeProxy` now has an explicit closing state:

- `close()` sets `closing = true` **before** destroying the tracked sockets;
- `track()` destroys any socket presented after that point instead of adding it to the (already drained) set;
- both request paths re-check `closing` immediately after their awaited `resolve()` and before creating/using the upstream connection — the HTTP path destroys the response, the CONNECT path destroys the client;
- `close()` is idempotent: the promise is memoized, so a repeated call awaits the first instead of asking a stopped server to close again.

Regression test is the injected lookup you described: it parks the CONNECT handler mid-await, `close()` starts, then the lookup is released. The upstream server counts accepted connections and the assertion is that it accepted **zero**, plus `close()` completing under the bound and a second `close()` succeeding. I checked it actually bites — removing either guard fails the test.

### 2. Real timeout normalization

Took the elapsed-deadline approach rather than message matching: the existing `name === 'TimeoutError' | 'AbortError'` branch stays, and `asFetchError` now also receives the deadline. Anything failing at or past it becomes the structured `TIMEOUT` envelope whatever the error calls itself; anything failing with budget left is returned untouched, so a refused connection or DNS failure can't be mislabelled.

Two tests: one rejecting with an impit-shaped generic `Error` (`operation timed out`, no special `name`) after the budget elapses, asserting `TIMEOUT`; one rejecting with `ECONNREFUSED` well inside a 30s budget, asserting the original error survives.

### Existing tests

Kept both — the idle CONNECT test and the ladder-throws test. Agreed neither substitutes for the delayed-lookup case.

### Security

DNS validation and every private-address check are unchanged; the closing state only ever *prevents* a connection, never permits one. `vitest run --project unit src/fetch` passes (26 tests), `tsc --noEmit` clean.

PR #266, opened 2026-08-10
fix: stop CSP allow-lists triggering challenge detection (#264)

### LINKED ISSUE #264: [Bug]: web fetch challenge detection false-positives on CSP headers (news.ycombinator.com fails with FETCH_BLOCKED)

ankitranjan7:
Fixes #264.

## The problem

`webcmd web fetch` wrongly decides a normal page is a bot-challenge page, and gives up on it.

The check that spots challenge pages searched **every response header** for words like `cloudflare`, `recaptcha` and `captcha`. Plenty of healthy sites list those domains in their `content-security-policy` header — that header is just an allow-list of third parties, not a statement about this response.

Hacker News is one of them. It returns a healthy `200` with the full page, gets labelled a challenge, retries through the whole impit ladder for nothing, and then fails:

```
$ webcmd web fetch --url https://news.ycombinator.com
CliError: The site blocked non-browser fetches.
  hint: Use webcmd web fetch-browser for this URL.
```

The page was never blocked. Any site whose CSP names a CDN or reCAPTCHA — a large share of the web — paid two wasted fetches and then a misleading `FETCH_BLOCKED`.

## What I changed

`src/fetch/classify.ts`:

1. Only look at headers that describe *this* response: `server`, `cf-mitigated`, `cf-chl-*`, `x-datadome*`, `set-cookie`. `content-security-policy`, `report-to` and `link` are ignored — they are third-party allow-lists and prove nothing.
2. A `200` now needs evidence in the **body** before it counts as a challenge. `server: cloudflare` on a page that was actually served is not a challenge; a real interstitial still says "Just a moment" in the body, and that still matches.
3. Body matching is unchanged.

## Proof

```
$ node dist/src/main.js web fetch --url https://news.ycombinator.com
# Hacker News   ← front page, plain tier, no ladder
```

- `npx vitest run src/fetch/` → 22 passed. `npm run typecheck` clean.
- New tests cover: the CSP false positive, a header-only `200`, and that a real `403` + `server: cloudflare` is still classified as a challenge (no under-detection).

### DIFF
--- a/src/fetch/classify.test.ts
+++ b/src/fetch/classify.test.ts
@@ -16,5 +16,14 @@ describe('fetch classification', () => {
     expect(isChallengeResponse(200, { server: 'cloudflare' }, '<html>real page</html>')).toBe(false);
     expect(isChallengeResponse(200, { server: 'cloudflare' }, 'Just a moment...')).toBe(true);
   });
+  it('treats cf-mitigated: challenge as decisive, even on a 200', () => {
+    expect(isChallengeResponse(200, { 'cf-mitigated': 'challenge' }, '<html>looks fine</html>')).toBe(true);
+    expect(isChallengeResponse(403, { 'cf-mitigated': 'challenge' }, 'forbidden')).toBe(true);
+  });
+  it('reads decisive evidence from the header name, not just its value', () => {
+    expect(isChallengeResponse(403, { 'cf-chl-out': 'AAAA1111' }, 'forbidden')).toBe(true);
+    expect(isChallengeResponse(403, { 'x-datadome': 'protected' }, 'forbidden')).toBe(true);
+    expect(isChallengeResponse(403, { 'x-datadome-cid': 'abc123' }, 'forbidden')).toBe(true);
+  });
   it('recognizes script-heavy app shells', () => expect(isJavaScriptShell('<div id="root"></div><script src="/app.js"></script><script>boot()</script>')).toBe(true));
 });
--- a/src/fetch/classify.test.ts
+++ b/src/fetch/classify.test.ts
@@ -5,6 +5,16 @@ describe('fetch classification', () => {
   it('recognizes explicit challenges but not bare forbidden responses', () => {
     expect(isChallengeResponse(403, { server: 'cloudflare' }, 'Just a moment...')).toBe(true);
     expect(isChallengeResponse(403, {}, 'forbidden')).toBe(false);
+    expect(isChallengeResponse(403, { server: 'cloudflare' }, 'forbidden')).toBe(true);
+  });
+  it('ignores third-party allow-lists in CSP and friends', () => {
+    const csp = "default-src 'self'; script-src https://www.google.com/recaptcha/ https://cdnjs.cloudflare.com/";
+    expect(isChallengeResponse(200, { 'content-security-policy': csp }, '<html>Hacker News</html>')).toBe(false);
+    expect(isChallengeResponse(403, { 'content-security-policy': csp }, 'forbidden')).toBe(false);
+  });
+  it('does not treat a served 200 as a challenge on headers alone', () => {
+    expect(isChallengeResponse(200, { server: 'cloudflare' }, '<html>real page</html>')).toBe(false);
+    expect(isChallengeResponse(200, { server: 'cloudflare' }, 'Just a moment...')).toBe(true);
   });
   it('recognizes script-heavy app shells', () => expect(isJavaScriptShell('<div id="root"></div><script src="/app.js"></script><script>boot()</script>')).toBe(true));
 });
--- a/src/fetch/classify.ts
+++ b/src/fetch/classify.ts
@@ -1,14 +1,22 @@
 const challengeMarkers = /cloudflare|cf-chl|datadome|perimeterx|px-captcha|akamai|captcha|just a moment|verify you are human/i;
-// Headers that say something about *this* response. CSP/report-to/link are allow-lists of third
-// parties (cdnjs.cloudflare.com, google.com/recaptcha) and are evidence of nothing.
-const signalHeaders = /^(?:server|cf-mitigated|cf-chl-[\w-]+|x-datadome[\w-]*|set-cookie)$/i;
+// Headers a bot-mitigation product sets *because it challenged this request*. The evidence is in
+// the name — values are opaque tokens — so these are decisive on their own, whatever the status.
+// https://developers.cloudflare.com/cloudflare-challenges/challenge-types/challenge-pages/detect-response/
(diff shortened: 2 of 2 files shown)


beubax:

Issue #264 is valid. A CSP or reporting header that merely allows `cdnjs.cloudflare.com` or Google reCAPTCHA is not evidence that the current response is a challenge. Restricting header inspection to response-specific signals, and refusing to classify a healthy `200` from a weak `server: cloudflare` header alone, is the right direction.

The current implementation drops some of the strongest challenge signals, however.

### Blocking classification gaps

The new code first filters by header name but then applies `challengeMarkers` only to the **value**:

https://github.com/agentrhq/webcmd/blob/b8bd78798c8ad75fcd5ad8fb4efa70f2b034b6b4/src/fetch/classify.ts#L1-L14

This loses evidence carried primarily in the header name:

- `cf-chl-*` with a neutral/non-marker value;
- `x-datadome*` with a neutral/non-marker value.

There is also a concrete Cloudflare regression: `cf-mitigated: challenge` is Cloudflare's documented machine-readable indicator for a Challenge Page, but the marker regex does not include the word `challenge`. The new code therefore fails to recognize that exact header/value pair. On status `200`, it returns false before considering any header-only decisive signal at all.

Cloudflare documents `cf-mitigated: challenge` as the reliable detection method here:
https://developers.cloudflare.com/cloudflare-challenges/challenge-types/challenge-pages/detect-response/

### Suggested correction

Keep the useful header allow-list, but distinguish decisive response headers from weak provider branding:

- inspect normalized `key:value` for `cf-chl-*` and `x-datadome*`, rather than discarding the key;
- treat `cf-mitigated: challenge` explicitly as decisive, including on `200`;
- keep `server: cloudflare` weak: it may support a blocked `403/429/503`, but must not turn an ordinary `200` into a challenge;
- continue ignoring CSP, report-to, link, and other third-party allow-lists.

A small table-driven classifier is enough; this does not need a larger abstraction.

### Required regression coverage

Please add cases for:

1. `200 + cf-mitigated: challenge` => challenge;
2. `403 + cf-chl-<suffix>: <neutral value>` => challenge;
3. `403 + x-datadome: <neutral value>` => challenge;
4. `200 + server: cloudflare + real body` => not a challenge;
5. `200/403 + CSP mentioning Cloudflare/reCAPTCHA + real/forbidden body` => not a challenge unless another decisive signal exists;
6. body markers such as “Just a moment” remain challenges.

### Scope

This classifier currently affects the client-owned `web fetch` path, including when the CLI is configured for hosted mode, so the fix requires a new CLI package but not a Cloud deployment. If `web fetch` later becomes server-owned, the same cases should be included in parity coverage.

The false-positive fix is worth landing, but the decisive-header regressions need to be closed first.

ankitranjan7:
Good catch on the `cf-mitigated` regression — pushed `34ffd00`.

**Classification now has three tiers instead of one filter:**

- **Decisive, evidence in the name** — `cf-chl-*` and `x-datadome*` are matched on the normalized header *name*, so a neutral/opaque value no longer discards the signal. `cf-mitigated: challenge` is handled explicitly (the value regex is `/challenge/i`, independent of `challengeMarkers`), per the Cloudflare detect-response doc you linked. These are checked **before** the `200` early return, so a healthy-looking `200` carrying `cf-mitigated: challenge` is now correctly a challenge.
- **Weak** — `server` and `set-cookie` still require a marker in the value, and are only consulted for `403/429/503`. `server: cloudflare` on a `200` with a real body stays not-a-challenge.
- **Ignored** — CSP, `report-to`, `link` and any other third-party allow-list are in neither list.

`challenge` was deliberately not added to the body/value `challengeMarkers` regex — it's too common in ordinary page text; it's only decisive as a `cf-mitigated` value.

**Regression coverage** — all six cases you listed:

1. `200 + cf-mitigated: challenge` → challenge ✅ (new)
2. `403 + cf-chl-out: AAAA1111` (neutral value) → challenge ✅ (new)
3. `403 + x-datadome: protected` and `x-datadome-cid: abc123` → challenge ✅ (new)
4. `200 + server: cloudflare + real body` → not a challenge ✅
5. `200/403 + CSP mentioning Cloudflare/reCAPTCHA` → not a challenge ✅
6. `Just a moment...` body → challenge ✅

`vitest run --project unit src/fetch` passes (24 tests), `tsc --noEmit` clean. Kept it to the table of name/value rules — no extra abstraction. Agreed on scope: client-owned `web fetch` path, new CLI package but no Cloud deploy.

beubax:
Thanks for the careful fix here. The CSP/header false positive was a valid bug: ordinary pages can mention Cloudflare/captcha/CDN providers in allow-list headers without being challenge pages. We landed the fix as part of #295, where `web fetch` moved into the core command path and the classifier now ignores unrelated headers like CSP/report-to/link while keeping challenge-specific evidence. Closing this PR as superseded by #295.

Issue #248, opened 2026-08-10
smart-search: mandatory SERP-first cost order wastes the budget when the query names a site


ankitranjan7:
### Problem

`skills/smart-search/SKILL.md` makes search-engine-first ordering non-negotiable:

> Cost order is mandatory: `webcmd web fetch` first, `webcmd web fetch-browser` second, search adapters last.

> For a search query without a direct URL, start with fetched search-engine result pages, not adapters.

> Do not jump to adapters because one engine blocked.

> Only after fetch-first search, target-page fetch, and allowed browser fetches fail or are insufficient, discover search adapters.

That ordering is right when the target site is unknown. It is backwards when the request names the sites. For "research browser automation on Hacker News and Reddit", the correct first move is HN Algolia and the `reddit` adapter — both purpose-built, structured, and unblocked.

### Observed cost

A real session following the skill correctly:

| Step | Outcome |
|---|---|
| DuckDuckGo HTML | empty body (302 + nginx) |
| Bing | CAPTCHA wall |
| `reddit.com/search.json` | WAF security page |
| `old.reddit.com/search` | wrong content — unrelated article via readability failure |
| `news.ycombinator.com/item?id=…` | "Sorry." block page |
| pullpush.io | `FETCH_BLOCKED` → escalate to `fetch-browser` → not installed (#247) |
| **HN Algolia API** | **worked — primary evidence** |
| **`webcmd reddit search --subreddit …`** | **worked — primary evidence** |

Everything that produced the final report was site-native. Everything the mandatory ordering tried first was blocked, and the skill explicitly forbids switching after one engine blocks.

### Proposed change

Add a site-native fast path ahead of the cost order: when the request names a known community or site (Hacker News, Reddit, Stack Overflow, GitHub, …), go to that site's adapter or public API first, and use SERPs only for discovery when no site is named or the named site has no adapter.

Supporting tweaks from the same session:

- Treat an empty extraction (0 usable result URLs) as a **failure**, not a success with no results.
- Treat a recognised block/challenge page as a hard blocker for that engine — do not re-fetch variants of it.
- Preflight adapter availability before reporting a site as unavailable; "Reddit is blocked" was reported while the `reddit` adapter simply was not installed.
- Warn on query tokens that collide with everyday English. `puppeteer` returned puppet videos and politics.

PR #271, opened 2026-08-10
fix: ship web fetch-browser so the blocked-fetch escalation works (#247)

### LINKED ISSUE #247: [Bug]: web fetch-browser is not installed by default, so the FETCH_BLOCKED escalation path is a dead end

ankitranjan7:
Fixes #247

## The problem

When a plain fetch is blocked, webcmd tells the agent to escalate to `webcmd web fetch-browser` — and on a fresh install that command does not work. The escalation path is a dead end.

The symptom has moved since the issue was filed. On 0.6.0 the generated `cli-manifest.json` **does** ship and advertise `web/fetch-browser`, so the command is now discoverable and instead dies when it runs (reproduced against a globally installed 0.6.0):

```
code: ADAPTER_LOAD
message: Cannot find module .../@agentrhq/webcmd/clis/web/fetch-browser.js
```

Same root cause, same dead end.

**Root cause:** [0470ac43](https://github.com/agentrhq/webcmd/commit/0470ac43) (#216) removed `"clis/"` from `package.json` `files` while moving adapters into plugins. `clis/web` went to neither place — it is not in the package and there is no `web` plugin either (`webcmd-plugin.json` registers 122 plugins; none is `web`), so `webcmd plugin search web` offers no way out.

## What I changed

Ship it in the core package — the issue's first option. The published package cannot carry adapter source at the root, because [check-package-bin.mjs:73](scripts/check-package-bin.mjs:73) fails the build on a top-level `clis/` or `plugins/` in the tarball, and that policy stays. So the build stages the tree next to the compiled output instead:

1. `scripts/copy-yaml.cjs` — copies `clis/` → `dist/src/clis/` (tests excluded), the same staging it already does for `playwright-client.js`. `files` already ships `dist/src/`.
2. `src/main.ts` — `BUILTIN_CLIS` uses the repo-root `clis/` when it exists (dev, `tsx src/main.ts`) and falls back to `<compiled dir>/clis` in an installed package.
3. `src/build-manifest.ts` — writes a second copy of the manifest to `dist/src/cli-manifest.json`, because the lookup contract is `clisDir/../cli-manifest.json`. Without it, an installed CLI falls back to a filesystem scan and eagerly imports every adapter at startup.
4. `scripts/check-package-bin.mjs` — new assertion: **every `modulePath` in `cli-manifest.json` must exist in the packed tarball.** That is exactly the regression this bug was, so it cannot come back silently.

## Proof

- `npm pack` → install the tarball into a temp prefix → `webcmd web fetch-browser --url https://example.com --stdout` returns the extracted Markdown. Before: `ADAPTER_LOAD`.
- `check:package-bin` passes. With the staging step reverted it fails with `packed tarball is missing manifest module for web/fetch-browser` — the new guard bites on the real regression.
- `cli-manifest.json` and `hosted-contract.json` regenerate byte-identical (`check:hosted-contract` passes). `check:plugin-parity`, `typecheck` and `npm test` (394 files, 4781 tests) all pass.

## Not in this PR

- **Reimplementing `fetch-browser` as a client-owned fast path next to `web fetch`.** The browser lifecycle it needs (page provisioning, session, cloak, pre-nav, observation, tab cleanup) lives in `execution.ts` around the adapter, so a literal port would duplicate all of it. This delivers what the core option promised — ships in the package, zero install steps, the pair works out of the box — with one browser lifecycle. Say the word if you want the literal port instead.

## Review notes

- No conflict with #263: that PR changes the *content* of `clis/web/fetch.js`, not its location. It also benefits — its discoverability entry for `web fetch` pointed at a file fresh installs did not contain, and now they do.
- #269 (`#248`, smart-search) is the other half of that session's wasted cost; different files, no overlap.

### DIFF
--- a/scripts/check-package-bin.mjs
+++ b/scripts/check-package-bin.mjs
@@ -4,7 +4,7 @@ import { spawnSync } from 'node:child_process';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import {
   formatPackageBinSpawnFailure,
   packageBinSpawnOptions,
@@ -112,6 +112,33 @@ try {
       fail(`global install did not create executable: ${binPath}`);
     }
     run(binPath, ['--version'], { cwd: tmp });
+    // Presence in the tarball is not executability. Every manifest command must
+    // also be discoverable and loadable *from the installed layout*, which a
+    // repository checkout never exercises: there the builtin tree is clis/ at
+    // the package root, and only the installed package takes the dist/src/clis
+    // fallback. #247 was exactly this gap — advertised, then ADAPTER_LOAD on
+    // first use.
+    const installedMain = path.join(prefix, 'lib', 'node_modules', pkg.name, pkg.main);
+    for (const entry of manifest) {
+      if (!entry.modulePath) continue;
+      const help = run(binPath, [entry.site, entry.name, '--help'], { cwd: tmp });
+      if (/ADAPTER_LOAD/.test(`${help.stdout}${help.stderr}`)) {
+        fail(`installed package cannot present ${entry.site}/${entry.name}: ADAPTER_LOAD\n${help.stdout}${help.stderr}`);
+      }
+      // --help is served from the staged manifest and never imports the module;
+      // the import below is what execution does, through the CLI's own resolver.
+      const load = [
+        `const { resolveBuiltinClisDir } = await import(${JSON.stringify(pathToFileURL(path.join(path.dirname(installedMain), 'package-paths.js')).href)});`,
+        `const { pathToFileURL } = await import('node:url');`,
+        `const { join } = await import('node:path');`,
+        `const dir = resolveBuiltinClisDir(${JSON.stringify(installedMain)});`,
+        `await import(pathToFileURL(join(dir, ${JSON.stringify(entry.modulePath)})).href);`,
+      ].join('\n');
+      const result = spawnSync(process.execPath, ['--input-type=module', '-e', load], { cwd: tmp, encoding: 'utf8' });
+      if (result.status !== 0) {
+        fail(`installed package cannot load ${entry.site}/${entry.name} (${entry.modulePath}) — this is the ADAPTER_LOAD failure users hit:\n${result.stderr}`);
+      }
+    }
   }
 } finally {
   fs.rmSync(tmp, { recursive: true, force: true });
--- a/scripts/check-package-bin.mjs
+++ b/scripts/check-package-bin.mjs
@@ -118,7 +118,15 @@ try {
     // the package root, and only the installed package takes the dist/src/clis
     // fallback. #247 was exactly this gap — advertised, then ADAPTER_LOAD on
     // first use.
-    const installedMain = path.join(prefix, 'lib', 'node_modules', pkg.name, pkg.main);
+    // npm -g --prefix puts packages under <prefix>/lib/node_modules on POSIX
+    // and directly under <prefix>/node_modules on Windows.
+    const installedRoot = process.platform === 'win32'
+      ? path.join(prefix, 'node_modules', pkg.name)
+      : path.join(prefix, 'lib', 'node_modules', pkg.name);
+    const installedMain = path.join(installedRoot, pkg.main);
+    if (!fs.existsSync(installedMain)) {
+      fail(`global install did not create the package entry point: ${installedMain}`);
+    }
     for (const entry of manifest) {
       if (!entry.modulePath) continue;
       const help = run(binPath, [entry.site, entry.name, '--help'], { cwd: tmp });
--- a/scripts/check-package-bin.mjs
+++ b/scripts/check-package-bin.mjs
@@ -78,6 +78,22 @@ try {
   if (packedPaths.has('scripts/fetch-adapters.js')) {
     fail('packed tarball contains the retired adapter fetch lifecycle');
   }
+
+  // Every command the core manifest advertises must be in the tarball. A
+  // manifest entry whose module was left behind is discoverable but dies with
+  // ADAPTER_LOAD on first use — see #247, where web/fetch-browser was the only
+  // escalation path FETCH_BLOCKED knew how to recommend.
+  const manifest = JSON.parse(fs.readFileSync(path.join(ROOT, 'cli-manifest.json'), 'utf8'));
+  for (const entry of manifest) {
+    if (!entry.modulePath) continue;
+    const packedModule = `dist/src/clis/${entry.modulePath}`;
+    if (!packedPaths.has(packedModule)) {
+      fail(`packed tarball is missing manifest module for ${entry.site}/${entry.name}: ${packedModule}`);
+    }
+  }
+  if (manifest.length > 0 && !packedPaths.has('dist/src/cli-manifest.json')) {
+    fail('packed tarball is missing the staged core manifest: dist/src/cli-manifest.json');
+  }
   for (const [name, target] of binEntries) {
     if (!packedPaths.has(String(target))) {
       fail(`packed tarball is missing bin "${name}" target: ${target}`);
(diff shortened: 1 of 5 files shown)


beubax:

The reported defect in #247 is valid, and the local packaging direction here is good. PR #216 removed the root `clis/` tree from the npm package while the generated manifest, `web fetch` error hint, and bundled smart-search skill continued to advertise `web/fetch-browser`. On a fresh 0.6.0 installation the command is therefore discoverable but fails at module load time.

Copying the small remaining built-in tree to `dist/src/clis`, staging the matching manifest beside it, and making `main.ts` fall back to that installed location is a reasonable minimal local-mode repair. It also correctly reuses the existing adapter execution lifecycle instead of duplicating browser provisioning in a second client-owned implementation.

There are two blockers before this is safe to merge.

### 1. The deployed Cloud loader cannot consume the new package layout

Webcmd Cloud currently checks only `<packageRoot>/clis`:

- [default adapter source detection](https://github.com/agentrhq/webcmd-cloud/blob/871132f1b4fe4c6871008d4f0b418617d1577a3a/src/default-adapters/source.ts#L274-L292)
- [default adapter execution resolution](https://github.com/agentrhq/webcmd-cloud/blob/871132f1b4fe4c6871008d4f0b418617d1577a3a/src/executor/load-command.ts#L327-L342)

When that root directory is absent, Cloud deliberately sets the default manifest to an empty array and later refuses to resolve default package modules. This PR ships only `dist/src/clis`; the new `BUILTIN_CLIS` resolution is private to `main.ts` and there is no exported package-layout resolver Cloud can call.

Consequently, publishing this change and bumping Cloud to that package would repair local installations but would still leave Cloud with zero default adapter commands. In particular, hosted `web fetch-browser` would remain hidden/unexecutable even though the packaged CLI works.

Please coordinate one of these before merge:

1. Update Cloud to discover the staged directory from the installed package entry point (for example, the `clis` sibling of the resolved `@agentrhq/webcmd` runtime), while retaining realpath/containment checks; or
2. Add a small public package-layout resolver only if the Cloud-side fallback proves insufficient.

The Cloud change needs contract and executability coverage, an exact `@agentrhq/webcmd` pin bump, and deployment sequencing. A marketplace/plugin reconciliation is not a substitute for loading the built-in package.

### 2. CI proves presence, not installed execution

The new `check-package-bin.mjs` assertion is valuable: it proves every manifest module path has a corresponding file in the tarball. However, the installed-tarball execution check still exercises only `webcmd --version`. It does not load `fetch-browser` through the new `dist/src/clis` fallback, which is the actual regression being fixed.

Please add a no-network installed-package smoke such as:

```sh
webcmd web fetch-browser --help
```

The assertion should verify successful command loading and explicitly guard against `ADAPTER_LOAD`. Running it from the temporary npm-installed prefix is important; a repository checkout does not exercise the fallback branch.

### Interaction with #263

If #263 is also retained, the Cloud work must explicitly decide how its client-owned `web/fetch` manifest entry is represented. Once Cloud starts reading the staged manifest, that command must not accidentally become a server-side default merely because it is `PUBLIC`.

### Merge condition

I am happy with the local implementation once:

- an installed-tarball command-load regression test is added; and
- the coordinated Cloud path change and tests are ready, or this PR is explicitly scoped and tracked as a local-only partial fix rather than closing hosted parity.

No new security defect is apparent in the OpenCLI patch itself; the Cloud resolver must preserve the existing symlink/realpath containment guarantees.

ankitranjan7:
Thanks for the detailed review. Pushed `b58d352` for blocker 2, and I'd like to take you up on the explicit-scoping option for blocker 1.

### 2. Installed execution, not just presence — done

You're right that `--version` proved nothing about the fallback. One correction on the suggested smoke, though: `webcmd web fetch-browser --help` does **not** exercise it. Help is rendered from the staged `dist/src/cli-manifest.json`; the module is only imported at execution time (`execution.ts`, where `adapterLoadError` is raised). I verified this by pointing the fallback at a non-existent directory — `--help` and `list` still succeeded.

So `check:package-bin` now does both, per manifest entry, from the temporary npm-installed prefix:

1. `<prefix>/bin/webcmd <site> <name> --help` — must exit 0 and must not mention `ADAPTER_LOAD` (discoverability through the staged manifest);
2. the module is then imported through the CLI's **own** resolver against the installed entry point — the exact import `execution.ts` performs, and the one that raised `ADAPTER_LOAD` in #247. No network, no browser launch.

To make (2) test the real resolution rather than a re-implementation that can drift, the `clis/`-at-root vs `dist/src/clis` decision moved out of `main.ts` into `resolveBuiltinClisDir()` in `package-paths.ts`. It is **not** added to the package `exports` map — per your note, no public resolver unless the Cloud-side fallback proves insufficient.

Mutation-checked: changing the fallback to a wrong directory fails the check with `ERR_MODULE_NOT_FOUND` on `dist/src/clis-nope/web/fetch-browser.js`. It passes on the real build. `tsc --noEmit` clean; full unit + plugin suite passes (4781 tests).

### 1. Cloud loader — requesting the explicit local-only scope

I can't land the `webcmd-cloud` resolver change from this repo, and I'd rather not hold the local repair behind a cross-repo deploy. Taking your second merge condition: **this PR is a local-only partial fix.** It repairs fresh local installs of `web fetch-browser`; it does **not** close hosted parity, and Cloud still sees zero default adapter commands until its loader learns the staged layout.

Concretely, Cloud work still owed and not covered here: discovering the staged directory from the resolved `@agentrhq/webcmd` entry point (retaining realpath/containment checks), contract + executability coverage, the exact package pin bump, and deployment sequencing. Happy to file that in `webcmd-cloud` and link it here — say the word and I'll open it, or point me at an existing tracking issue. If you'd rather this PR not merge until the Cloud side is ready, that's fine too; the local fix is self-contained either way.

### Interaction with #263

Noted — and #263 is where I'll encode that ownership decision rather than letting it fall out of packaging. I've replied there proposing the client-owned contract (option 1), so `web/fetch` is excluded from the hosted manifest by metadata and doesn't become a server-side default once Cloud starts reading the staged manifest.

No change to the symlink/realpath containment guarantees here.

ankitranjan7:
Follow-up `d081bb9`: the new load check resolved the installed package root as `<prefix>/lib/node_modules/…`, which is POSIX-only — `check:package-bin` also runs on the Windows runner, where `npm -g --prefix` installs to `<prefix>/node_modules/…`. Now branched on platform, with an explicit failure if the entry point is missing rather than a confusing `ERR_MODULE_NOT_FOUND` later.

beubax:
Thanks for chasing the missing `web fetch-browser` install path — the reported failure was real. After discussing the architecture, we chose a different resolution in #295: `web fetch` is now an inbuilt/client-owned core command, and `web fetch-browser`/`clis/web` were removed instead of shipped as default adapters. Browser rendering is now an explicit Session + `browser run` workflow after fetch fails, not an automatic/missing adapter escalation. Closing this PR as superseded by #295.

PR #278, opened 2026-08-11
fix(browser): serialize undefined run results as null

rishabhraj36:
## Description

Related issue:

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

- [ ] I updated tests or docs if needed
- [ ] I included output or screenshots when useful

### Adapter Notes

### DIFF
--- a/src/browser/run/runner.test.ts
+++ b/src/browser/run/runner.test.ts
@@ -405,6 +405,14 @@ describe('runBrowserProgram', () => {
     expect(output.result).toBe('undefined');
   });
 
+  it('serializes a program with no explicit return as null', async () => {
+    const output = await run(`
+      await page.getByRole('button', { name: 'Save' }).click();
+    `);
+
+    expect(output.result).toBeNull();
+  });
+
   it('keeps time, memory, output, redaction, and serialization limits', async () => {
     await expect(run('while (true) {}', { timeoutMs: 25 })).rejects.toMatchObject({
       code: 'BROWSER_RUN_TIMEOUT',
--- a/src/browser/run/runner.ts
+++ b/src/browser/run/runner.ts
@@ -446,7 +446,7 @@ export async function runBrowserProgram(
             rejectRun = undefined;
           }
           try {
-            const serialized = JSON.stringify(value);
+            const serialized = JSON.stringify(value === undefined ? null : value);
             if (serialized === undefined) throw new TypeError('Result is not JSON serializable.');
             return serialized;
           } catch (cause) {


beubax:

The one-line behavior change is sensible, but this PR needs a regression test and a minimal problem statement before it is safe to merge.

### Why the behavior is valid

`new AsyncFunction(source)()` returns JavaScript `undefined` when a browser program performs side effects but has no explicit `return`. `JSON.stringify(undefined)` itself returns `undefined`, so the current runner reports `BROWSER_RUN_SERIALIZATION_ERROR` even though the program executed successfully.

Representing that top-level “no value” as JSON `null` is a reasonable stable-envelope convention. The proposed expression preserves explicit `null` and leaves genuinely unserializable values—such as `BigInt`—on the existing error path.

### Why a test is required

This branch changes the public result contract for both local and hosted browser runs:

- local Cloak calls `runBrowserProgram` directly;
- Webcmd Cloud imports `runBrowserProgram` from the pinned `@agentrhq/webcmd/browser/run` export.

A later cleanup could easily revert `undefined` to the existing serialization error while all current tests remain green, because no test invokes a program without a return value. The existing serialization test covers `BigInt`, not the normal no-return case.

Please add a focused test in `src/browser/run/runner.test.ts` equivalent to:

```ts
it('serializes a program with no explicit return as null', async () => {
  const output = await run(`
    await page.locator('button').click();
  `);

  expect(output.result).toBeNull();
});
```

Keep or explicitly assert the complementary invariant:

```ts
await expect(run('return 1n;')).rejects.toMatchObject({
  code: 'BROWSER_RUN_SERIALIZATION_ERROR',
});
```

A CLI/envelope assertion would also be useful if there is already a cheap browser-run fixture, but one runner-level regression test is the minimum needed.

### PR context

The description currently contains only the repository template, has no reproduction/output, and links no issue. Please document:

- the failing no-return source;
- the current `BROWSER_RUN_SERIALIZATION_ERROR`;
- the intended `result: null` output;
- whether this was observed locally, through Cloud, or both.

### Deployment implication

Once merged and released, local users receive the behavior in the new CLI package. Cloud receives it only after its exact `@agentrhq/webcmd` pin is bumped and deployed. The test matters for maintaining parity across those two consumers.

With that small test and description update, I would consider the implementation appropriate and minimal.

rishabhraj36:
> ## Maintainer review: changes requested
> The one-line behavior change is sensible, but this PR needs a regression test and a minimal problem statement before it is safe to merge.
> 
> ### Why the behavior is valid
> `new AsyncFunction(source)()` returns JavaScript `undefined` when a browser program performs side effects but has no explicit `return`. `JSON.stringify(undefined)` itself returns `undefined`, so the current runner reports `BROWSER_RUN_SERIALIZATION_ERROR` even though the program executed successfully.
> 
> Representing that top-level “no value” as JSON `null` is a reasonable stable-envelope convention. The proposed expression preserves explicit `null` and leaves genuinely unserializable values—such as `BigInt`—on the existing error path.
> 
> ### Why a test is required
> This branch changes the public result contract for both local and hosted browser runs:
> 
> * local Cloak calls `runBrowserProgram` directly;
> * Webcmd Cloud imports `runBrowserProgram` from the pinned `@agentrhq/webcmd/browser/run` export.
> 
> A later cleanup could easily revert `undefined` to the existing serialization error while all current tests remain green, because no test invokes a program without a return value. The existing serialization test covers `BigInt`, not the normal no-return case.
> 
> Please add a focused test in `src/browser/run/runner.test.ts` equivalent to:
> 
> ```ts
> it('serializes a program with no explicit return as null', async () => {
>   const output = await run(`
>     await page.locator('button').click();
>   `);
> 
>   expect(output.result).toBeNull();
> });
> ```
> 
> Keep or explicitly assert the complementary invariant:
> 
> ```ts
> await expect(run('return 1n;')).rejects.toMatchObject({
>   code: 'BROWSER_RUN_SERIALIZATION_ERROR',
> });
> ```
> 
> A CLI/envelope assertion would also be useful if there is already a cheap browser-run fixture, but one runner-level regression test is the minimum needed.
> 
> ### PR context
> The description currently contains only the repository template, has no reproduction/output, and links no issue. Please document:
> 
> * the failing no-return source;
> * the current `BROWSER_RUN_SERIALIZATION_ERROR`;
> * the intended `result: null` output;
> * whether this was observed locally, through Cloud, or both.
> 
> ### Deployment implication
> Once merged and released, local users receive the behavior in the new CLI package. Cloud receives it only after its exact `@agentrhq/webcmd` pin is bumped and deployed. The test matters for maintaining parity across those two consumers.
> 
> With that small test and description update, I would consider the implementation appropriate and minimal.

Added the runner-level regression test for programs without an explicit return. It asserts result: null, while the existing BigInt assertion continues to cover BROWSER_RUN_SERIALIZATION_ERROR.

PR #285, opened 2026-08-11
fix: tokenize plugin search queries instead of matching one literal substring

### LINKED ISSUE #282: plugin search: multi-word queries return nothing ("hacker news" misses hackernews)
`webcmd plugin search` matches the query as one literal substring, so any multi-word query misses.

## Repro (webcmd 0.6.0, node v22.23.1, macOS)

```bash
webcmd plugin search "hacker news" -f json   # => {"plugins": [], "errors": []}
webcmd plugin search hackernews -f json      # => hackernews plugin
webcmd plugin search hacker -f json          # => hackernews plugin
```

## Why it matters

`skills/webcmd-usage/SKILL.md` sends agents down this exact path: registry filter → empty → `webcmd plugin search <query>` → empty → raw browser fallback. Agents naturally derive the query from the user's words ("pull the top 5 stories from Hacker News" → `hacker news`), get zero results, and conclude no adapter exists. The `hackernews` plugin is right there. In our run of `start.md` this cost the whole verification step: we fell through to `webcmd web fetch` instead of installing the adapter.

Note the skill even tells agents to "preserve the user's term when practical", which makes the spaced form more likely, not less.

## Suggested fix

Tokenize the query and match tokens independently against the plugin name/description, ignoring separators — so `hacker news`, `hacker-news`, and `hackernews` all hit. Cheapest version: strip non-alphanumerics from both query and haystack before the substring test, then AND the remaining tokens.

Worth doing the same for `webcmd list -f json` consumers if that path shares the matcher.

Agnik47:
## Description

`searchCatalogPlugins` tested the whole query as one literal substring of `"name description"`, so multi-word queries missed a one-word plugin name:

```
$ webcmd plugin search "hacker news" -f json   # => {"plugins": [], "errors": []}
$ webcmd plugin search hackernews -f json      # => hackernews plugin
$ webcmd plugin search hacker -f json          # => hackernews plugin
```

This matters because `skills/webcmd-usage/SKILL.md` sends agents down registry filter → empty → `plugin search <query>` → empty → raw browser fallback, and agents naturally derive multi-word queries from the user's own words ("pull the top 5 stories from Hacker News" → `hacker news`). A real adapter (`hackernews`) exists and is invisible to the exact query an agent is most likely to try.

### Fix

Strip non-alphanumeric separators from both the query and the `"name description"` haystack, split the query into independent tokens, and require every token to match (AND semantics) rather than testing the query as one contiguous string. This makes `hacker news`, `hacker-news`, and `hackernews` all match, and makes token order irrelevant (`news hacker` matches too), while a query like `reddit news` still correctly finds nothing.

I checked `webcmd list -f json` (mentioned in the issue as worth checking) — it filters by `--tag`, not a free-text query, and doesn't call `searchCatalogPlugins` at all, so it doesn't share this matcher and needs no change.

Fixes #282

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

`npm run typecheck` and `npm run build` pass, generated artifacts stay byte-identical.

Reproduced the exact issue repro against the real bundled catalog, after the fix:

```
$ webcmd plugin search "hacker news" -f json
{
  "plugins": [
    { "name": "hackernews", "description": "Webcmd commands for hackernews", ... }
  ],
  "errors": []
}
```

Added a test to `src/plugin-catalog.test.ts` covering `hacker news`, `hacker-news`, `hackernews`, `hacker`, and reordered `news hacker` all matching, plus `reddit news` correctly matching nothing:

```
$ npx vitest run --project unit src/plugin-catalog.test.ts
 Test Files  1 passed (1)
      Tests  8 passed (8)
```


ankitranjan7:
Thank you @Agnik47 for the thorough implementation and original regression coverage. Before merging, we incorporated the name-focused fixture from #284 and the separated all-terms assertion from #288 so the final change reflects the strongest ideas from all three contributions. Clean verification on current main passed: build succeeded and 5,065 tests passed. Squash-merging this as the canonical fix for #282.

PR #286, opened 2026-08-11
fix: handle unhandled EPIPE crash in web fetch proxy, and Cloudflare 200 false positive

### LINKED ISSUE #283: web fetch broken: unhandled EPIPE crash in safe-proxy, plus FETCH_BLOCKED false positive on Cloudflare-fronted 200s
`webcmd web fetch` currently fails on both of the URLs we tried, in two different ways. Environment: webcmd 0.6.0 (fresh `webcmd update`), node v22.23.1, macOS 15 (Darwin 25.6.0), `webcmd doctor` fully green (daemon 0.6.0, cloak 0.4.5 connected).

---

## 1. Unhandled EPIPE crashes the CLI

```bash
webcmd web fetch --url https://news.ycombinator.com/
```

```
node:events:497
      throw er; // Unhandled 'error' event
      ^
Error: This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (node:net:576:14)
    at Socket.ondata (node:internal/streams/readable:1009:22)
    ...
  code: 'EPIPE'
}
```

Reproducible 3/3, before and after `webcmd daemon restart`. The process dies with a raw Node stack — no `CliError`, no error envelope, no JSON — so an agent gets nothing structured to act on.

Root cause looks like `dist/src/fetch/safe-proxy.js:54-68` (the CONNECT handler). The tunnel is wired as:

```js
upstream.once('connect', () => { ...; upstream.pipe(client); client.pipe(upstream); });
upstream.once('error', error => client.destroy(error));
```

There is no `error` listener on `client` at all, and the `upstream` listener is `once`, so a write to a half-closed peer emits an unhandled `'error'` and takes down the process. Related: `proxy.close()` only calls `server.close()`, which does not destroy established tunnel sockets, so in-flight tunnels race the teardown at the end of `webFetch`.

Suggested fix: persistent `client.on('error', …)` / `upstream.on('error', …)` that destroy the peer, and track + destroy open sockets in `close()`. Same missing-handler issue applies to the plain HTTP path (`request.pipe(upstream)` at line 48).

---

## 2. `FETCH_BLOCKED` false positive on any Cloudflare-fronted 200

```bash
webcmd web fetch --url https://example.com
# CliError: The site blocked non-browser fetches.  code: 'FETCH_BLOCKED'
```

`example.com` is a plain static 200 with no challenge:

```
HTTP/2 200
content-type: text/html
server: cloudflare
cf-cache-status: HIT
```

`dist/src/fetch/classify.js` builds its evidence string from **headers plus body** and tests `/cloudflare|…/i` against it, gated on `status === 200` among others. Any origin behind Cloudflare emits `server: cloudflare` and `cf-*` headers on perfectly good 200 responses — which today is a large fraction of the web — so the classifier reports a block that did not happen, burns both impit retries, and then pushes the agent to `fetch-browser` for a page plain HTTP already returned.

Suggested fix: don't treat header *values* naming a CDN as challenge evidence. Match challenge markers against the body (and specific challenge headers like `cf-mitigated`), and drop `status === 200` from the gate unless a body marker such as "Just a moment" is actually present.

---

Both were hit while following `start.md` step 3 (the Hacker News verification task) on a clean install.

Agnik47:
## Description

Two independent bugs in `web fetch`, both hit on a clean install following `start.md` step 3.

### 1. Unhandled EPIPE crashes the whole CLI process

`safe-proxy.ts`'s CONNECT tunnel handler registered an error listener on the upstream leg only:

```js
upstream.once('connect', () => { ...; upstream.pipe(client); client.pipe(upstream); });
upstream.once('error', error => client.destroy(error));
```

There was no `client.on('error', ...)` at all. Once the client-facing socket resets or half-closes mid-tunnel, relaying more data to it throws an unhandled `'error'` event with no listener — Node's default behavior for that is to crash the entire process. The plain HTTP path (`request.pipe(upstream)`) had the same gap on `request`/`response`.

Fix: persistent, symmetric error handlers on both legs of both paths (CONNECT tunnel and plain HTTP), each destroying its peer instead of leaking the error.

Related: `proxy.close()` only called `server.close()`, which merely stops accepting *new* connections — per Node's docs, its callback doesn't fire until every existing connection ends on its own, and an established CONNECT tunnel never does that by itself. So `close()` would hang indefinitely with a tunnel in flight, racing whatever `webFetch()` does next. Now the proxy tracks every socket/request it opens and destroys them before closing the server.

I verified this is real and not just a "should be fine" default: I reproduced the exact reported crash locally (`Error: read ECONNRESET` → unhandled `'error'` event → process exit 1) against the pre-fix code, in an isolated child process, before writing the fix.

### 2. `FETCH_BLOCKED` false positive on any Cloudflare-fronted 200

`classify.ts` built its evidence string from headers *and* body, and matched `/cloudflare|.../i` against it, gated on `status === 200` among other statuses:

```js
const challengeMarkers = /cloudflare|cf-chl|datadome|perimeterx|px-captcha|akamai|captcha|just a moment|verify you are human/i;
return challengeMarkers.test(evidence) && (status === 403 || status === 429 || status === 503 || status === 200);
```

Any response fronted by Cloudflare emits `server: cloudflare` and `cf-*` headers on perfectly ordinary 200s — which today is a large fraction of the web — so the classifier reported a block that never happened, burned both impit retries, and then pushed the agent to `fetch-browser` for a page plain HTTP already returned successfully.

Fix: split the markers into two tiers. Strings that don't legitimately appear except on an actual challenge page (`cf-chl`, `cf-mitigated`, `datadome`, `perimeterx`, `px-captcha`, `captcha`, `just a moment`, `verify you are human`) stay decisive at any status, including 200 — this is what still catches a genuine Cloudflare managed-challenge interstitial served with a 200. A bare CDN/vendor name (`cloudflare`, `akamai`) is no longer decisive on its own; it now only corroborates an already-suspicious non-200 status (403/429/503).

Fixes #283

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

`npm run typecheck` and `npm run build` pass, generated artifacts stay byte-identical.

Reproduced the exact `example.com` repro from the issue against the real bundled build, after the fix:

```
# Example Domain
...
```
(before the fix: `CliError: The site blocked non-browser fetches.  code: 'FETCH_BLOCKED'`)

Added regression tests for both bugs, and — since it's easy to write a test that passes by construction rather than actually catching the bug — verified each one fails against the original code before the fix:

- `src/fetch/classify.test.ts`: the Cloudflare-200 case fails against pre-fix `classify.ts` (`expected true to be false`), and still catches a genuine challenge (403/503 with a vendor name, or a 200 with a body/header challenge marker).
- `src/fetch/safe-proxy.test.ts`: the crash repro runs in an isolated child process (a crashed process would otherwise take the whole vitest worker down with it, not just fail one test) and asserts exit code 0 with no `ECONNRESET` in stderr — against pre-fix code it reliably reproduces the exact reported crash. A second test confirms `close()` no longer hangs on a still-open tunnel.

```
$ npx vitest run --project unit src/fetch/
 Test Files  5 passed (5)
      Tests  26 passed (26)
```

Full `unit` project also passes, no new failures beyond the same pre-existing local-environment-only ones present on unmodified `main`.

### DIFF
--- a/src/fetch/classify.test.ts
+++ b/src/fetch/classify.test.ts
@@ -6,5 +6,19 @@ describe('fetch classification', () => {
     expect(isChallengeResponse(403, { server: 'cloudflare' }, 'Just a moment...')).toBe(true);
     expect(isChallengeResponse(403, {}, 'forbidden')).toBe(false);
   });
+  it('does not flag an ordinary 200 just because it is served through a CDN (issue #283)', () => {
+    expect(isChallengeResponse(200, { server: 'cloudflare', 'cf-cache-status': 'HIT', 'content-type': 'text/html' }, '<html><body>Example Domain</body></html>')).toBe(false);
+    expect(isChallengeResponse(200, { server: 'akamaighost' }, '<html>ordinary page</html>')).toBe(false);
+  });
+  it('still flags a CDN vendor name as a challenge signal once the status itself looks blocked', () => {
+    expect(isChallengeResponse(503, { server: 'cloudflare' }, '')).toBe(true);
+    expect(isChallengeResponse(429, { server: 'akamaighost' }, '')).toBe(true);
+  });
+  it('flags a 200 challenge interstitial via a body marker, without needing a CDN header', () => {
+    expect(isChallengeResponse(200, {}, 'Just a moment...')).toBe(true);
+  });
+  it('flags a 200 challenge via a specific challenge header even without body markers', () => {
+    expect(isChallengeResponse(200, { server: 'cloudflare', 'cf-mitigated': 'challenge' }, '<html>interstitial</html>')).toBe(true);
+  });
   it('recognizes script-heavy app shells', () => expect(isJavaScriptShell('<div id="root"></div><script src="/app.js"></script><script>boot()</script>')).toBe(true));
 });
--- a/src/fetch/classify.ts
+++ b/src/fetch/classify.ts
@@ -1,8 +1,19 @@
-const challengeMarkers = /cloudflare|cf-chl|datadome|perimeterx|px-captcha|akamai|captcha|just a moment|verify you are human/i;
+// Strings that don't legitimately appear except on an actual challenge/block page —
+// decisive on their own, at any status including a plain 200 (e.g. a Cloudflare
+// managed challenge can render its interstitial with a 200).
+const strongChallengeMarkers = /cf-chl|cf-mitigated|datadome|perimeterx|px-captcha|captcha|just a moment|verify you are human/i;
+// Bare CDN/vendor names are not evidence of blocking by themselves — `server: cloudflare`
+// (or akamai) shows up on every response those networks front, challenged or not, so
+// treating it as decisive on a 200 flags most of the ordinary CDN-fronted web. Only
+// corroborate an already-suspicious non-200 status with it.
+const cdnVendorMarkers = /cloudflare|akamai/i;
 
 export function isChallengeResponse(status: number, headers: Record<string, string>, body: string): boolean {
   const evidence = `${Object.entries(headers).map(([key, value]) => `${key}:${value}`).join('\n')}\n${body.slice(0, 20_000)}`;
-  return challengeMarkers.test(evidence) && (status === 403 || status === 429 || status === 503 || status === 200);
+  const blockedStatus = status === 403 || status === 429 || status === 503;
+  if (strongChallengeMarkers.test(evidence)) return blockedStatus || status === 200;
+  if (cdnVendorMarkers.test(evidence)) return blockedStatus;
+  return false;
 }
 
 export function isJavaScriptShell(body: string): boolean {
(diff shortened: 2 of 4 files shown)


beubax:
Thanks for the very solid #283 investigation. We used the important parts of this work in #295: the Cloudflare-fronted healthy-200 classifier fix, plus the safe-proxy reset/error handling idea so tunnel peers are torn down instead of surfacing EPIPE/ECONNRESET crashes. #295 also removed the old browser-escalation adapter split, so the final shape differs from this PR, but the bug report and proxy/classifier analysis were genuinely useful. Closing this PR as superseded by #295.

PR #287, opened 2026-08-11
fix: harden safe-proxy CONNECT tunnels and stop flagging CDN 200s as blocked

### LINKED ISSUE #283: web fetch broken: unhandled EPIPE crash in safe-proxy, plus FETCH_BLOCKED false positive on Cloudflare-fronted 200s

ayushsingh82:
## Summary
- Persistent `error` handlers on both the client and upstream sockets in the CONNECT tunnel (plus tracked-socket teardown in `close()`) so a write to an already half-closed peer no longer crashes the process with an unhandled `EPIPE`, and `close()` no longer hangs on in-flight tunnels.
- `isChallengeResponse` now matches challenge markers only against the response body (plus a small set of challenge-specific header names like `cf-mitigated`), so a plain Cloudflare-fronted 200 no longer trips `FETCH_BLOCKED`.

Fixes #283

## Test plan
- [x] `npx tsc --noEmit`
- [x] Added regression tests in `src/fetch/safe-proxy.test.ts` (tunnel reset doesn't crash the process; `close()` tears down in-flight tunnels instead of hanging) and `src/fetch/classify.test.ts` (Cloudflare-fronted 200 with a normal body is not flagged); verified both fail against the pre-fix code and pass after.

### DIFF
--- a/src/fetch/classify.test.ts
+++ b/src/fetch/classify.test.ts
@@ -7,4 +7,13 @@ describe('fetch classification', () => {
     expect(isChallengeResponse(403, {}, 'forbidden')).toBe(false);
   });
   it('recognizes script-heavy app shells', () => expect(isJavaScriptShell('<div id="root"></div><script src="/app.js"></script><script>boot()</script>')).toBe(true));
+  it('does not flag a Cloudflare-fronted 200 with a normal body', () => {
+    expect(isChallengeResponse(200, { server: 'cloudflare', 'cf-cache-status': 'HIT' }, '<html><body>Hello world</body></html>')).toBe(false);
+  });
+  it('flags a 200 whose body shows an actual challenge page', () => {
+    expect(isChallengeResponse(200, { server: 'cloudflare' }, 'Just a moment...')).toBe(true);
+  });
+  it('flags a challenge header on a non-200 status even without body evidence', () => {
+    expect(isChallengeResponse(403, { 'cf-mitigated': 'challenge' }, 'forbidden')).toBe(true);
+  });
 });
--- a/src/fetch/classify.ts
+++ b/src/fetch/classify.ts
@@ -1,8 +1,17 @@
-const challengeMarkers = /cloudflare|cf-chl|datadome|perimeterx|px-captcha|akamai|captcha|just a moment|verify you are human/i;
+// Matched only against the response body: CDN *names* (e.g. "server: cloudflare") are not
+// evidence of a challenge, since most of the web sits behind one on perfectly good responses.
+const bodyChallengeMarkers = /cf-chl|datadome|perimeterx|px-captcha|akamai.*captcha|captcha|just a moment|checking your browser|verify you are human|attention required/i;
+// Header *names* that only ever appear when a challenge actually fired.
+const challengeHeaderNames = ['cf-mitigated', 'x-datadome-captcha', 'x-px-block'];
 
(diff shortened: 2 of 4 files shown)


ayushsingh82:
Fixes #283 — @ngaurav this resolves both the EPIPE crash and the FETCH_BLOCKED false positive. Ready for review.

beubax:
Thanks for hardening the same #283 area. The concern was valid, and #295 now covers it in the broader `web fetch` core-command refactor: healthy CDN-fronted 200s are no longer treated as blocked, and safe-proxy tears down peer sockets on reset/error to avoid the EPIPE/ECONNRESET crash class. Since #295 is merged with the final architecture, closing this PR as superseded.

PR #288, opened 2026-08-11
fix: tokenize plugin search queries so multi-word terms match

### LINKED ISSUE #282: plugin search: multi-word queries return nothing ("hacker news" misses hackernews)

ayushsingh82:
## Summary
- `searchCatalogPlugins` matched the query as one literal substring, so a multi-word query like `hacker news` missed the `hackernews` plugin even though `hackernews` and `hacker` alone both matched.
- The query is now split into tokens, non-alphanumerics are stripped from both tokens and the name/description haystack, and every token must match — so `hacker news`, `hacker-news`, and `hackernews` all hit.

Fixes #282

## Test plan
- [x] `npx tsc --noEmit`
- [x] Added regression tests in `src/plugin-catalog.test.ts` covering the exact repro from the issue and confirming multi-token queries still require every token to match; verified they fail against the pre-fix code and pass after.

### DIFF
--- a/src/plugin-catalog.test.ts
+++ b/src/plugin-catalog.test.ts
@@ -119,4 +119,28 @@ describe('plugin catalog', () => {
     expect(result.plugins.map((plugin) => plugin.name)).toEqual(['flights']);
     expect(result.errors).toEqual([{ sourceId: 'bad', manifestUrl: 'https://bad.test/webcmd-plugin.json', message: 'network failed' }]);
   });
+
+  it('matches multi-word queries against a single-word plugin name, ignoring separators', async () => {
+    const catalog: PluginCatalog = {
+      version: 1,
+      sources: [{ id: 'ok', source: 'github:ok/repo', manifestUrl: 'https://ok.test/webcmd-plugin.json' }],
+    };
+    const fetchJson = async () => ({ plugins: { hackernews: { path: 'plugins/hackernews', description: 'Browse Hacker News stories and comments' } } });
+
+    for (const query of ['hacker news', 'hacker-news', 'hackernews', 'hacker']) {
+      const result = await searchCatalogPlugins(catalog, { query, fetchJson });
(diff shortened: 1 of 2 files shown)


ayushsingh82:
Fixes #282 — @ngaurav multi-word plugin search (e.g. "hacker news") now resolves correctly. Ready for review.

ankitranjan7:
Thank you @ayushsingh82 for the thoughtful implementation and clear regression coverage. We merged #285 as the canonical implementation to avoid duplicate changes, but incorporated your separation of the positive matching case from the all-query-terms-required negative case. You are also credited as a co-author on the squash commit. We really appreciate the contribution and hope you will contribute again!

PR #289, opened 2026-08-11
fix: detect skill/CLI version drift between npm and Claude Code plugin installs

### LINKED ISSUE #274: Skill/CLI version drift between npm and Claude Code plugin install paths
## Why this issue exists

Webcmd ships skills through more than one install path. A user can update the npm CLI while still running older skills from an agent plugin/cache path.

This issue predates the 2026-08-20 eval issues. It is still relevant because skill text affects agent behavior, and stale skill text can make agents call commands or flags that no longer match the installed CLI.

Related PR:

- #289: `fix: detect skill/CLI version drift between npm and Claude Code plugin installs` is still open.

## Current Webcmd state

- `webcmd skills add/update/remove/list` exists for linked skill installs.
- Linked skills are refreshed by `webcmd update` unless `--skip-skills` is passed.
- Codex and Claude plugin manifests also bundle `skills/` directly.
- Docs now warn not to install both plugin-bundled skills and `webcmd skills add` copies.
- Claude docs say plugin updates are version-gated and separate from `webcmd update`.
- There is still no merged drift detector in `webcmd doctor` for plugin-cache skills versus the running CLI.

## Observed agent/user behavior

The risky state is:

1. User installs Webcmd as an agent plugin.
2. Plugin copies skills into an agent-owned versioned cache.
3. User later runs `webcmd update` and upgrades only the npm CLI.
4. The agent still reads older plugin-bundled skill text.
5. The skill can recommend stale commands, flags, or workflows against the newer CLI.

This is especially costly because agents trust skill text as authoritative.

## What went wrong

The system treats CLI version and plugin-bundled skill version as separate update surfaces, but the user-facing mental model is one product: Webcmd.

Docs reduce the chance of duplicate installs, but they do not detect or repair an already-drifted environment.

## Solution directions

Product-first options:

1. Add a `webcmd doctor` drift check that reports the running CLI version, linked skill version, and plugin-cache skill version when detectable.
2. Make the warning actionable: which command refreshes which install path.
3. Keep `webcmd update` behavior simple. It can refresh linked skills, but plugin-cache updates may need to stay with the host agent's plugin manager.
4. Consider a docs/contributing note that skill changes require a release-triggering commit type so plugin users receive them.

- Should `webcmd update` try to invoke `claude plugin update`, or should it only diagnose and print instructions?
- Which plugin caches can be detected reliably across Claude, Codex, and other agents?
- Is #289 the right implementation shape, or should the scope shrink to `doctor` only?

## Acceptance criteria

- `webcmd doctor` can identify a likely CLI/skill version drift state.
- The diagnostic distinguishes linked skills from plugin-bundled cached skills.
- The message tells the user exactly how to refresh the stale side.
- The fix does not duplicate skills or mutate agent plugin caches unexpectedly.

## Non-goals

- No automatic writes into third-party plugin caches unless explicitly chosen.
- No broad agent-plugin manager abstraction.
- No requirement that npm and every agent plugin update through the same command.

ayushsingh82:
## Summary
Webcmd ships `skills/` through two channels: an npm symlink (always in sync with the running package) and a Claude Code plugin-cache copy that's version-pinned and only refreshed by `claude plugin update`. Since only the CLI's own version nags on update-check, a user can end up running a newer CLI against older skills with no signal, producing confidently wrong commands.

- Adds `findSkillVersionDrift()`: when both channels are present on disk and the plugin-cache version differs from the running CLI, `webcmd doctor` now reports it with the fix command (`claude plugin update webcmd@webcmd`). This is item 2 from the issue — the cheapest check that catches the actual broken state.
- Adds a CONTRIBUTING.md note (item 3): `docs:`/`chore:` commits don't cut a release-please version, so a skill fix landed under those prefixes never reaches plugin users even though npm installs pick it up for free.

Item 1 (`webcmd update` passthrough to `claude plugin update`) is left for a follow-up since it's more invasive (shelling out to another CLI) and this PR already covers the two lower-risk, high-value pieces.

Fixes #274 (items 2 and 3)

## Test plan
- [x] `npx tsc --noEmit`
- [x] Added `src/skill-version-drift.test.ts` covering: drift reported when both channels present and versions differ, no drift when versions match, no drift when only one channel is installed, and picks the highest of several installed plugin-cache versions.
- [x] `src/doctor.test.ts` still passes unchanged (23/23) — the new field doesn't affect existing report assertions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,14 @@ import { CommandExecutionError, EmptyResultError } from '@agentrhq/webcmd/errors
 
 User adapters, plugins, cache, traces, and site memory live under `~/.webcmd`.
 
+## Skill Changes and Release Triggers
+
+The Claude Code plugin install path copies `skills/` into a version-pinned cache directory
+and only refreshes it when release-please cuts a new version. `docs:`/`chore:` commits do not
+cut a version, so a skill fix landed under those prefixes stays invisible to plugin users
+indefinitely even though npm installs (which symlink `skills/`) pick it up immediately. Use
+`fix:`/`feat:` for any commit that changes a file under `skills/`.
+
 ## Documentation
 
 The published docs at [webcmd.dev/docs](https://webcmd.dev/docs) are built by Mintlify from the `docs/` directory in this repo. To change the published docs, edit the `.mdx` pages under `docs/` (and `docs/docs.json` for navigation) — do not edit the site directly.
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -14,6 +14,7 @@ import type { BrowserProfileStatus } from './browser/daemon-transport.js';
 import { aliasForContextId, loadProfileConfig } from './browser/profile.js';
 import { formatDaemonVersion, isDaemonStale, staleDaemonIssue } from './browser/daemon-version.js';
 import { findShadowedUserAdapters, formatAdapterShadowIssue, type AdapterShadow } from './adapter-shadow.js';
+import { findSkillVersionDrift, formatSkillVersionDriftIssue, type SkillVersionDrift } from './skill-version-drift.js';
 
 const DOCTOR_LIVE_TIMEOUT_SECONDS = 8;
 const DOCTOR_SESSION = '__doctor__';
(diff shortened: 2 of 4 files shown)


ayushsingh82:
Fixes #274 (items 2 and 3) — @ankitranjan7 adds the doctor drift check plus the CONTRIBUTING.md note. Ready for review.

PR #290, opened 2026-08-11
fix: give each profile runtime an unleased anchor page

### LINKED ISSUE #276: [Bug]: releasing the final leased page can invalidate the shared profile runtime
## Summary

In the persistent local Webcmd daemon, releasing or closing the final leased page can cause Chromium to close its browser context. A command arriving immediately afterward may reuse the runtime before the asynchronous context close event invalidates it, then fail on its first navigation with:

```text
Target page, context or browser has been closed
```

This is a Webcmd runtime lifecycle race, independent of benchmark isolation, network quality, model behavior, the serialization issue, CloakBrowser session limits in #225, and the profile-prefix teardown bug in #242.

## Current behavior

`CloakSessionManager.getPage()` adopts the first existing Chromium page when the profile has no tracked pages:

```ts
const existingPages = candidate.context.pages();
return !freshPage && existingPages[0] && candidate.pages.size === 0
  ? existingPages[0]
  : this.createPage(candidate.context, input.windowMode);
```

That initial `about:blank` page therefore becomes an ordinary leased user page.

When the session is released, Webcmd removes and closes every matching non-persistent page:

```ts
if (entry.siteSession !== "persistent" && !pageIsClosed(entry.page)) {
  await entry.page.close().catch(() => {});
}
```

If this is the final Chromium page, `page.close()` can begin browser/context shutdown. The call can resolve before Chromium shutdown propagation completes. During that interval:

1. The old `ProfileRuntime` is still present in `profiles`.
2. A new command obtains that runtime.
3. It creates or adopts a page.
4. The old context finishes closing.
5. The new command fails during `goto()` with the generic target-closed error.

Waiting briefly between release and the next command avoids the failure because the context `close` event has time to run `invalidateProfileRuntime()`. Immediate reuse reproduces it.

The same invariant can be violated through `release()`, explicit `closePage()`, `freshPage`, and idle-timeout cleanup.

## Expected behavior

Ending a task/session should release its user pages without implicitly shutting down the profile runtime owned by the persistent daemon. A subsequent command should be able to create and navigate a page immediately.

Separate profiles must remain independent and usable concurrently. For example, these should work without restarting the daemon:

```bash
webcmd --profile peer twitter login
webcmd --profile actor twitter login
webcmd --profile peer twitter whoami
webcmd --profile actor twitter whoami
```

## Suggested fix: runtime-owned anchor page

Give each `ProfileRuntime` one unleased `about:blank` anchor page:

- Reserve the initial Chromium page as `runtime.anchorPage` when launching the profile runtime.
- Never lease it to a task or site session.
- Exclude it from `listPages`, tab selection, page IDs, snapshots, and network capture.
- Always create a separate page for the first user lease.
- Preserve the anchor through `release()`, `closePage()`, `freshPage`, and idle cleanup.
- If the anchor is unexpectedly closed while the context is healthy, recreate it under the existing per-profile page-creation lock before closing the final leased page.
- Close the anchor only when intentionally shutting down or evicting that profile runtime, shutting down the daemon, or after an unrecoverable browser disconnect.
- For `freshPage`, create and register the replacement before closing the previous leased page.

The anchor must be per profile, not global. The manager already maintains a `Map<string, ProfileRuntime>`, and each profile has a separate user-data directory, so `peer` and `actor` would each have their own anchor and login state.

## Acceptance criteria

- Repeated `release final session -> immediately create page -> navigate` cycles do not produce target-closed errors.
- Closing the final visible/leased page leaves the profile context connected.
- The anchor never appears in user-facing tab/page APIs.
- `freshPage` cannot transiently close the final runtime-owned page.
- Two named profiles can remain active and switch back and forth without a daemon restart.
- Shutting down the daemon closes every profile context and anchor.
- Intentional per-profile idle eviction closes only that profile runtime; persisted cookies remain available after relaunch.
- Tests cover explicit close, release, fresh-page replacement, idle cleanup, immediate reuse, and multiple profiles.

## Additional note: window mode

Keeping an anchor prevents unnecessary background/foreground relaunches during normal daemon operation, but it does not by itself define behavior after a real daemon/browser restart or when an existing background profile receives a foreground login request. Canonical per-profile window-mode selection/promotion should be handled separately rather than coupling it to this lifecycle fix.

ayushsingh82:
## Summary
`CloakSessionManager.getPage()` adopted the context's initial `about:blank` page as an ordinary leased user page the first time a session asked for one. Releasing that final leased page could then begin browser/context shutdown; a command arriving before that shutdown finished propagating would fail on its first navigation with `Target page, context or browser has been closed`.

Implements the anchor-page approach from the issue:
- The context's initial page is reserved as `ProfileRuntime.anchorPage` at launch (adopted synchronously, no extra page creation on the common path) instead of being adopted into the pages map — so it's automatically excluded from `listPages`, tab selection, page IDs, and network capture, since those all iterate the pages map.
- Every `getPage()`/`newPage()` lease now always creates its own page — no more adopting a leftover tab.
- If the anchor was never created (the practically nonexistent case of a context launching with zero pages) or gets unexpectedly closed while the context is still healthy, it's recreated — under the existing per-profile page-creation lock — before the next leased-page close, so closing the last leased page can never drop the context to zero pages.
- The anchor is per `ProfileRuntime`, so separate profiles (e.g. `peer` / `actor`) stay independent.

Fixes #276

## Test plan
- [x] `npx tsc --noEmit`
- [x] Added a `runtime-owned anchor page` describe block to `session-manager.test.ts` covering: the anchor stays open when the only leased page is released, immediate reuse after release doesn't relaunch the context, the anchor is recreated if unexpectedly closed before the next page close, the anchor is never exposed via `listPages`, each profile gets its own independent anchor, and `freshPage` never leases the anchor out.
- [x] Updated 3 pre-existing tests whose expected `newPage`/CDP call counts changed now that the first lease always creates its own page instead of adopting the anchor (this is the actual behavior fix, not a test artifact — verified by reverting the fix and confirming these tests, old and new, fail as expected).
- [x] Updated `provider.test.ts`'s shared page-mock fixture to model the anchor correctly (own initial tab, never leased); fixed 2 tests whose hardcoded page-array indices shifted because the first lease now creates a real second page instead of adopting the anchor.
- [x] Full local run of `src/browser/**/*.test.ts` plus the existing `unit` project: no failures beyond the pre-existing, unrelated `src/hosted/*` baseline and two files that need a locally installed Chromium binary this sandbox doesn't have.

### DIFF
--- a/src/browser/runtime/local-cloak/provider.test.ts
+++ b/src/browser/runtime/local-cloak/provider.test.ts
@@ -47,17 +47,29 @@ function fakePage(url: string, initialViewport: { width: number; height: number 
 }
 
 function makeProviderWithFakePage(initialViewport: { width: number; height: number } | null = { width: 1280, height: 720 }) {
-  const pages = [fakePage('https://example.com/', initialViewport)];
+  // `anchor` stands in for Chromium's own default initial tab — the runtime reserves it as
+  // ProfileRuntime.anchorPage on launch and never leases it out. `page` is the object the
+  // first real newPage() call hands back (whether triggered by a session's first getPage or
+  // an explicit new tab), matching what every test here treats as "the" first real page —
+  // pages[0] stays the untouched anchor, pages[1] is the first real one, exactly like the
+  // pre-fix layout, so index-based assertions on multi-tab flows don't need to change.
+  const anchor = fakePage('about:blank');
+  const page = fakePage('https://example.com/', initialViewport);
+  const pages = [anchor];
+  let realPagesCreated = 0;
   const cdpSession = { send: vi.fn().mockResolvedValue(undefined), detach: vi.fn().mockResolvedValue(undefined) };
   const browser = { contexts: vi.fn(() => [context]) };
   const context = {
     browser: vi.fn(() => browser),
     on: vi.fn(),
-    pages: vi.fn(() => pages.filter((page) => !page.isClosed())),
+    pages: vi.fn(() => pages.filter((p) => !p.isClosed())),
     newPage: vi.fn(async () => {
-      const page = fakePage('about:blank');
-      pages.push(page);
-      return page;
+      // A counter, not pages.length — some tests push externally-simulated pages (e.g. a
+      // popup) into `pages` before any real newPage() call happens.
+      const created = realPagesCreated === 0 ? page : fakePage('about:blank');
+      realPagesCreated += 1;
+      pages.push(created);
+      return created;
     }),
     newCDPSession: vi.fn().mockResolvedValue(cdpSession),
     cookies: vi.fn().mockResolvedValue([{ name: 'sid', value: '1', domain: 'example.com', path: '/' }]),
@@ -67,7 +79,7 @@ function makeProviderWithFakePage(initialViewport: { width: number; height: numb
     baseDir: '/tmp/webcmd-test',
     launchPersistentContext: vi.fn().mockResolvedValue(context),
   });
-  return { provider, browser, page: pages[0], pages, context, cdpSession };
+  return { provider, browser, page, pages, context, cdpSession };
 }
 
 describe('LocalCloakRuntimeProvider', () => {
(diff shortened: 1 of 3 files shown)


ayushsingh82:
Fixes #276 — @rishabhraj36 @beubax implements the anchor-page approach from the issue. Ready for review.

ayushsingh82:
@ankitranjan7 plz have a look at this as well

ankitranjan7:
Thanks for the implementation. I’m closing this because the underlying lifecycle problem has since been addressed on `main` by #291, with additional first-window handling in #298. Current `main` now maintains a per-profile hidden anchor target, with a parking-page fallback and anchor recovery, so this older implementation is no longer needed.

PR #296, opened 2026-08-12
docs: fix Cloak Chromium pin

beubax:
## Summary
- Correct the documented Chromium artifact for cloakbrowser 0.4.5 to 145.0.7632.109.2
- Align the profile sessions spec and implementation plan with the live gate/runtime pin

## Tests
- rtk proxy git diff --check main...codex/fix-cloak-pin-docs


beubax:
Closing per direction: superpowers docs should stay local and not land on main.

PR #298, opened 2026-08-13
fix(browser): repair doctor session and first cloak window

beubax:
## Summary
- create a temporary opaque Session for `webcmd doctor` connectivity probes instead of the old `__doctor__` sentinel
- reuse Cloak's fresh launch `about:blank` page for the first Session window, avoiding an extra empty window after all Cloak windows are closed

## Verification
- `npx vitest run src/doctor.test.ts src/browser/runtime/local-cloak/session-manager.test.ts`
- `npm run typecheck`
- `npx vitest run src/browser/sessions.test.ts src/browser/runtime/local-cloak/provider.test.ts src/daemon/server.test.ts src/browser.test.ts`
- `npm test`
- `npm run build`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text eol=lf
+*.png binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,1 @@
 * text eol=lf
-*.png binary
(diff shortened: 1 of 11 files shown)


ankitranjan7:
Follow-up in `8f5fc66`:

- `webcmd doctor` now discards its temporary Session record after closing the runtime, preventing idle Session entries from accumulating.
- Normal explicit Sessions remain reusable; adapter-default and human-handoff Sessions are protected from discard.
- Added `*.png binary` to `.gitattributes` so PNGs are not modified by LF normalization.

Verification:
- Focused Session/doctor tests: 107 passed
- Full suite: 5,065 passed
- Typecheck and build passed

PR #301, opened 2026-08-13
Enforce plugin registration in root webcmd-plugin.json

### LINKED ISSUE #222: webcmd-adapter-author skill: root webcmd-plugin.json step is documented but skippable — should be enforced, not optional

ayushsingh82:
## Summary
- Fixes #222 — the root-manifest registration step in the `webcmd-adapter-author` skill's promotion flow could be silently skipped: `webcmd plugin create`, `webcmd plugin install`, and `webcmd validate <site>` all succeed even when a promoted plugin never lands in the root `webcmd-plugin.json`.
- `webcmd validate` (untargeted) now warns when a `plugins/<site>` directory has its own `webcmd-plugin.json` but is missing from the root manifest's `plugins` map, via a new `findUnregisteredPlugins()` in `plugin-manifest.ts`. No-op outside a monorepo checkout (e.g. npm installs, which ship neither `plugins/` nor a root manifest).
- `adapter-template.md` gets an explicit grep-based verification step right after the root-manifest edit, so an agent following the skill has to confirm the step landed instead of trusting that every command succeeding means it worked.

## Test plan
- [x] `npx vitest run src/plugin-manifest.test.ts src/validate.test.ts` — new coverage for `findUnregisteredPlugins` (missing/registered/disabled/no-monorepo/no-`plugins/`-dir cases) and the `webcmd validate` wiring
- [x] Manually verified against this repo: removed `amazon`'s entry from the root manifest, confirmed `webcmd validate` flagged it, restored and confirmed clean again
- [x] Full `unit` vitest project (147 files / 2356 tests) passes with these changes in the tree

### DIFF
--- a/skills/webcmd-adapter-author/references/adapter-template.md
+++ b/skills/webcmd-adapter-author/references/adapter-template.md
@@ -34,14 +34,23 @@ Then add `<site>` to the root `webcmd-plugin.json` `plugins` map:
 }
 ```
 
+This step is required, not optional — nothing in `webcmd plugin create`, `webcmd plugin install`, or `webcmd validate <site>` fails if it's skipped, so confirm it actually landed:
+
+```bash
+grep -q '"<site>"' webcmd-plugin.json || echo "MISSING: <site> not registered in root webcmd-plugin.json"
+```
+
 Before handing off, remove the private shadow and prove the plugin path works:
 
 ```bash
 rm -rf ~/.webcmd/clis/<site>
 webcmd plugin install file://$PWD/plugins/<site>
 webcmd validate <site>
+webcmd validate
 webcmd <site> <command> --help
 ```
+
+The bare `webcmd validate` (no target) also warns if `plugins/<site>` exists but was never added to the root `webcmd-plugin.json` `plugins` map.
 
 ## Minimal Registry Shape
 
--- a/src/plugin-manifest.test.ts
+++ b/src/plugin-manifest.test.ts
@@ -12,6 +12,7 @@ import {
   _getEnabledPlugins as getEnabledPlugins,
   _parseVersion as parseVersion,
   _satisfiesRange as satisfiesRange,
+  _findUnregisteredPlugins as findUnregisteredPlugins,
   MANIFEST_FILENAME,
   validatePluginAuthor,
   type PluginManifest,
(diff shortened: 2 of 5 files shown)


ankitranjan7:
Thanks for the thorough implementation and tests. I’m closing this because it enforces a workflow that is no longer intended: the root plugin catalog and README table are generated after merge from each plugin’s local manifest by the community-plugin sync workflow. A validation warning here would incorrectly flag contributions that follow the current generated-catalog process. #117 will align the authoring documentation with that workflow.

PR #302, opened 2026-08-13
Retry background Chromium launch via lsregister on stale LaunchServices cache

### LINKED ISSUE #220: webcmd doctor: browser connectivity check fails with kLSNoExecutableErr on a valid Chromium binary (stale LaunchServices cache), no auto-remediation
## What's wrong

`webcmd doctor` can fail its browser connectivity check even when the managed Chromium binary under `~/.cloakbrowser/<version>/Chromium.app` is completely intact and executable. The failure is a macOS LaunchServices (LS) database desync, not a broken install, and `doctor` doesn't detect or remediate it.

## Repro / evidence

```
$ webcmd doctor
[OK] Daemon: running on port 9777 (v0.5.2)
[OK] Runtime: cloak connected (v0.4.5)
[FAIL] Connectivity: failed (Command failed: /usr/bin/open -g -n /Users/ng/.cloakbrowser/chromium-145.0.7632.109.2/Chromium.app --args ... about:blank
The application cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-10827 "kLSNoExecutableErr: The executable is missing" ...)
```

But the binary itself is fine:
- `file Chromium.app/Contents/MacOS/Chromium` → `Mach-O 64-bit executable arm64`
- Running it directly (`.../Chromium --version`) → prints `Chromium 145.0.7632.109` and exits 0
- `codesign -dv` → valid adhoc signature, no quarantine xattr

The failure is specifically in `/usr/bin/open`, which resolves `.app` bundles via the LaunchServices database rather than executing the path directly. `kLSNoExecutableErr` here means LS's cached record for that bundle path is stale/out of sync with the filesystem — not that the executable is actually missing.

## Fix that works today (manual)

```
/System/Library/Frameworks/CoreServices.framework/Frameworks/LaunchServices.framework/Support/lsregister -f "/path/to/Chromium.app"
```

This forces LS to re-scan and re-register the bundle. After running it, `open` on the same bundle succeeds immediately and `webcmd doctor` passes.

## Why this likely happens

`~/.cloakbrowser/<version>/Chromium.app` is downloaded/unpacked by webcmd's own tooling (a script unzipping into a hidden dotfile directory), not installed via Finder/a `.pkg`. LaunchServices only reliably picks up new/changed bundles through specific triggers (background `lsd` rescans, Spotlight/`mdimporter` events, or an explicit `lsregister` call) — a script-driven unzip into a dotfile path can land outside those triggers, especially on a version bump/update where a previous bundle occupied a related path. In this instance, a `.last_update_check` file in the same directory was timestamped ~26 minutes before the failure, consistent with an update having just touched that tree.

## Suggested fix

`webcmd doctor` (or the underlying connectivity check) should detect `kLSNoExecutableErr` specifically and either:
- Automatically run `lsregister -f` on the managed Chromium bundle and retry once, or
- At minimum, print that exact remediation command in the `Issues:` section instead of just reporting a generic connectivity failure.

## Context

Hit this at the very start (Step 1: `webcmd doctor`) of a subagent building a new university-course-export adapter. It cost ~16 tool calls / ~1m45s of manual diagnosis (checking file validity, direct execution, code signature, then discovering `lsregister -f` was the fix) before recon could even start — and turned out to be moot in this case, since the adapter ultimately used `Strategy.PUBLIC` with `browser: false` and never needed the browser at all. Filed alongside #217, #218, #219 from the same session.

ayushsingh82:
## Summary
- Fixes #220 — `webcmd doctor`'s browser connectivity check can fail with `kLSNoExecutableErr` even though the managed Chromium bundle under `~/.cloakbrowser/<version>/Chromium.app` is intact and executable (verified: valid codesign, no quarantine xattr, runs fine when invoked directly). The failure is a stale macOS LaunchServices cache entry for the bundle path — likely because the bundle is unzipped into a hidden dotfile directory by webcmd's own tooling rather than installed via Finder/a `.pkg`, which can land outside the triggers that make LS pick it up.
- `launchDarwinBackgroundPersistentContext` (`darwin-background-launch.ts`) now detects `kLSNoExecutableErr` from the `open` failure, runs `lsregister -f` on the bundle, and retries the launch once.
- If the retry still fails, the thrown error includes the exact manual remediation command (`lsregister -f "<bundle>"`) instead of the generic connectivity failure text — this flows through to `webcmd doctor`'s `Issues:` section automatically since it already surfaces the underlying error message verbatim.
- Any other `open` failure (unrelated to a stale LS cache) is untouched — no remediation attempt, same behavior as before.

## Test plan
- [x] `npx vitest run src/browser/runtime/local-cloak/darwin-background-launch.test.ts` — new coverage: successful retry after re-registering, remediation error surfaced when retry still fails, unrelated failures skip remediation entirely (regression guard)
- [x] Full `unit` vitest project (147 files / 2356 tests) passes with this change in the tree

### DIFF
--- a/src/browser/runtime/local-cloak/darwin-background-launch.test.ts
+++ b/src/browser/runtime/local-cloak/darwin-background-launch.test.ts
@@ -30,6 +30,7 @@ function fakeDependencies(browser: Browser) {
     connectOverCDP: vi.fn().mockResolvedValue(browser),
     terminateProfile: vi.fn().mockResolvedValue(undefined),
     removePortFile: vi.fn().mockResolvedValue(undefined),
+    registerBundle: vi.fn().mockResolvedValue(undefined),
   };
 }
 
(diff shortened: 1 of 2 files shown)


ayushsingh82:
@ankitranjan7 plz let me know your review on this

ankitranjan7:
Thanks @ayushsingh82 — reviewed and merged.

PR #303, opened 2026-08-13
Distinguish missing browser binary from generic connectivity failure in doctor

### LINKED ISSUE #239: [Bug]: webcmd doctor reports generic "fetch failed" when CloakBrowser binary download is blocked
## Summary

`webcmd doctor` reports a generic `fetch failed` error when the CloakBrowser Chromium binary is missing and its automatic download cannot reach the download servers.

The daemon and Cloak runtime appear healthy, which makes the error look like a browser-profile or daemon problem rather than a missing browser binary.

## Environment

- OS: Fedora Linux 42
- Node.js: v20.20.2
- Webcmd: v0.5.3
- CloakBrowser: v0.4.5
- Installation: global npm package

## Workaround

Using an existing local Chrome installation resolves the problem:

```bash
export CLOAKBROWSER_BINARY_PATH=/opt/google/chrome/chrome
webcmd daemon restart
webcmd doctor -v
```

After applying the workaround:

```text
[OK] Daemon: running on port 9777 (v0.5.3)
[OK] Runtime: cloak connected (v0.4.5)

Profiles:
  • default: connected v0.4.5
[OK] Connectivity: connected in 0.1s

Everything looks good!
```

External browser navigation and the Reddit adapter also worked afterward.

## Suggested improvement

1. Check for the CloakBrowser executable during `webcmd doctor`.
2. Preserve the original binary download error and attempted URL.
3. Return a specific installation/download error code instead of `fetch failed` or `UNKNOWN`.
4. Suggest `CLOAKBROWSER_BINARY_PATH` as an actionable fallback.
5. Clarify that "runtime connected" does not mean the browser binary is installed and launchable.

## Additional observation

Invoking the installed CloakBrowser CLI through its npm symlink produced no output:

```bash
node_modules/.bin/cloakbrowser doctor
```

Invoking its actual script directly worked:

```bash
node node_modules/cloakbrowser/dist/cli.js info --quick --json
```

This appears related to the CLI entry-point check comparing `import.meta.url` with the unresolved symlink path. It may be a separate CloakBrowser issue, but it made diagnosis harder.

### Steps to Reproduce

Start with no CloakBrowser binary installed:

```bash
ls ~/.cloakbrowser
```

The directory is empty.

Use a network where the CloakBrowser and GitHub download endpoints cannot be reached, then run:

```bash
webcmd daemon restart
webcmd profile list
webcmd doctor -v
```

A browser-backed adapter also reproduces it:

```bash
webcmd reddit login
```

## Expected behavior

Webcmd should preserve and display the underlying CloakBrowser error, for example:

```text
CloakBrowser Chromium is not installed and could not be downloaded.

Download failed from cloakbrowser.dev and GitHub.
Check network access or set CLOAKBROWSER_BINARY_PATH to a compatible local Chromium executable.
```

The error code should ideally distinguish browser installation/download failures from page fetch failures, such as:

```text
BROWSER_BINARY_MISSING
BROWSER_BINARY_DOWNLOAD_FAILED
```

`webcmd doctor` could also check whether the expected CloakBrowser binary exists before attempting the browser connectivity probe.

## Actual behavior

`webcmd profile list` reports:

```text
No Cloak runtime profiles are active.
Run a browser-backed command or webcmd <site> login to create one.
```

`webcmd doctor -v` reports:

```text
webcmd v0.5.3 doctor (node v20.20.2)

[OK] Daemon: running on port 9777 (v0.5.3)
[OK] Runtime: cloak connected (v0.4.5)
[FAIL] Connectivity: failed (fetch failed)

Issues:
  • Browser connectivity test failed: fetch failed
```

`webcmd reddit login` reports:

```yaml
ok: false
error:
  code: UNKNOWN
  message: fetch failed
  exitCode: 1
```

The browser binary expected by CloakBrowser was not installed:

```json
{
  "binary": {
    "path": "~/.cloakbrowser/chromium-146.0.7680.177.5/chrome",
    "installed": false
  }
}
```

### webcmd Version

v0.5.3

### Node.js Version

20.x

### Operating System

Linux

### Logs / Screenshots

```shell

```

ayushsingh82:
## Summary
- Fixes #239 — `webcmd doctor` reports a generic `fetch failed` when the CloakBrowser Chromium binary is missing and its automatic download can't reach the download servers. The daemon and Cloak runtime report healthy, which makes it look like a browser-profile or daemon problem rather than a missing binary — the reporter's workaround was setting `CLOAKBROWSER_BINARY_PATH` to an existing local Chrome install after a lot of manual diagnosis.
- New `checkBrowserBinary()` in `doctor.ts` checks binary presence ahead of the live connectivity probe, via CloakBrowser's own `binaryInfo()` (already a public export of the `cloakbrowser` package webcmd depends on), or — when `CLOAKBROWSER_BINARY_PATH` is set — that override path directly, so a correctly-configured workaround isn't misreported as "missing."
- `doctor` now renders a distinct `Browser binary:` status line, separate from `Runtime:` — "runtime connected" only reflects the daemon/Cloak process, not whether the Chromium binary it needs to launch is actually installed.
- When connectivity fails specifically because the binary isn't installed, the `Issues:` section now reports the binary path, download URL, and the underlying error, with `CLOAKBROWSER_BINARY_PATH` suggested as an actionable fallback — instead of the old generic `Browser connectivity test failed: fetch failed`.

Scoped to `webcmd doctor` only. I did not touch the general adapter-command error-code path (e.g. `webcmd reddit login`'s `error.code: UNKNOWN`) — that funnels through the daemon IPC/retry-classification system (`session-manager.ts`, `daemon-client.ts`) and is a larger, separate change if wanted.

## Test plan
- [x] `npx vitest run src/doctor.test.ts` — new coverage: binary-installed case leaves the generic message untouched, binary-missing case produces the specific diagnostic (path/URL/underlying error/`CLOAKBROWSER_BINARY_PATH` hint, and confirms the old generic message does *not* also appear), `CLOAKBROWSER_BINARY_PATH` override is treated as authoritative over the managed cache state, override-pointing-nowhere case gets override-specific guidance, plus direct render tests for the new `Browser binary:` line in both states
- [x] Full `unit` vitest project (147 files / 2356 tests) passes with this change in the tree

### DIFF
--- a/src/doctor.test.ts
+++ b/src/doctor.test.ts
@@ -1,4 +1,7 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const {
   mockGetDaemonHealth,
@@ -51,6 +54,12 @@ vi.mock('./adapter-shadow.js', async () => {
 
 import { checkBrowserBinary, checkConnectivity, renderBrowserDoctorReport, runBrowserDoctor } from './doctor.js';
 
+const managedBinaryDir = fs.mkdtempSync(path.join(os.tmpdir(), 'webcmd-managed-binary-'));
+const managedBinaryPath = path.join(managedBinaryDir, process.platform === 'win32' ? 'chrome.exe' : 'chrome');
+fs.writeFileSync(managedBinaryPath, '#!/bin/sh\n');
+if (process.platform !== 'win32') fs.chmodSync(managedBinaryPath, 0o755);
+afterAll(() => fs.rmSync(managedBinaryDir, { recursive: true, force: true }));
+
 describe('doctor report rendering', () => {
   const strip = (s: string) => s.replace(/\x1b\[[0-9;]*m/g, '');
 
@@ -66,7 +75,7 @@ describe('doctor report rendering', () => {
       bundledVersion: '146.0.7680.177.5',
       tier: 'free',
       platform: 'linux-x64',
-      binaryPath: '/home/test/.cloakbrowser/chromium-146.0.7680.177.5/chrome',
+      binaryPath: managedBinaryPath,
       installed: true,
       cacheDir: '/home/test/.cloakbrowser/chromium-146.0.7680.177.5',
       downloadUrl: 'https://cloakbrowser.dev/chromium-v146.0.7680.177.5/cloakbrowser-linux-x64.tar.gz',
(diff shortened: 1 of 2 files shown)


adikulkarni006 (approved):

ayushsingh82:
@ankitranjan7 plz let me know your review on this

ankitranjan7:
Thanks @ayushsingh82 — reviewed and merged.

Issue #173, opened 2026-07-27
Correct skill and help guidance for format and verbose flags


rishabhraj36:
Parent: #171

## Problem

Skills call `-f/--format` and `-v/--verbose` universal even though many built-ins reject them. Some adapter-author references instruct agents to run an unsupported browser option.

Verified documentation gaps:

- `skills/webcmd-usage/SKILL.md` labels both flags universal.
- Several references use `webcmd browser network --format json`, but `browser network` rejects `--format`.
- Those browser examples also need the required `<session>` positional.
- Structured help supports `--help -f yaml|json` independently from runtime output formatting, which is not clearly distinguished.

Affected references include:

- `skills/webcmd-usage/SKILL.md`
- `skills/webcmd-adapter-author/references/site-recon.md`
- `skills/webcmd-adapter-author/references/api-discovery.md`
- `skills/webcmd-adapter-author/references/field-decode-playbook.md`

## Scope

- Replace “Universal Flags” with accurate wording based on implemented support.
- State that runtime `-f`/`-v` availability is command-specific and visible in command help.
- Document structured-help formatting separately.
- Fix browser examples to use valid flags and named-session syntax.
- Update guidance as the functionality sub-issues land.

## Acceptance criteria

- [ ] Every documented command example succeeds against the current CLI.
- [ ] Skills no longer claim unsupported flags are universal.
- [ ] Runtime formatting and structured-help formatting are clearly distinguished.
- [ ] Browser examples include `<session>` and no unsupported `--format` flag.

PR #315, opened 2026-08-14
fix: make -v effective in hosted, browser, and auth commands

### LINKED ISSUE #174: Make verbose diagnostics effective across hosted, browser, and auth commands
Parent: #171

## Problem

`-v/--verbose` is documented as universal, but support is inconsistent and hosted adapter execution currently accepts the flag without using it.

Verified gaps:

- Hosted adapter parsing accepts `-v`, but dispatch drops `parsed.verbose` (`src/hosted/runner.ts:269-307`).
- The hosted client execution contract has no verbose field (`src/hosted/client.ts:103-114`).
- Browser leaves reject `-v`, although CDP diagnostics already read `WEBCMD_VERBOSE` (`src/browser/cdp.ts:119-123`, `354-359`).
- `auth status` and `auth refresh` perform browser/daemon probes but reject `-v`.
- `doctor` already supports `-v`.

## Scope

- Forward hosted adapter verbose mode and make its effect observable, or reject it instead of silently ignoring it.
- Expose/apply verbose mode for browser leaves that perform bridge/CDP I/O.
- Exclude filesystem-only `browser init`.
- Decide explicitly whether `browser verify -v` forwards to its adapter subprocess.
- Add verbose support to auth probes only if useful underlying diagnostics are wired through.
- Do not add no-op verbose flags to list-only, filesystem-only, completion, acknowledgement-only, or passthrough commands.

## Acceptance criteria

- [ ] Hosted `-v` is effective or explicitly unsupported.
- [ ] Browser `-v` enables existing bridge/CDP diagnostics.
- [ ] Any auth `-v` support has observable diagnostics.
- [ ] No command advertises a no-op verbose flag.
- [ ] Local and hosted behavior is covered by tests.

Agnik47:
Closes #174.

## Problem

`-v/--verbose` was advertised more widely than it worked.

The real no-op was hosted mode. `parseCommandSurface` produces `parsed.verbose`, but `dispatchHosted` passed only `format`/`trace`/`profile`/`session` to `client.execute` and dropped it:

```ts
// src/hosted/runner.ts — before
: await client.execute({
    command: command.command,
    args: parsed.args,
    format: parsed.format,
    trace: parsed.trace,
    profile: parsed.profile ?? normalized.profile,
    session: normalized.session,
  });
```

So `webcmd <site> <cmd> -v` did nothing in the cloud while the identical flag worked locally (`commanderAdapter.ts` sets `WEBCMD_VERBOSE`). `HostedClient` also had no verbose diagnostics anywhere, so forwarding the flag alone would still have printed nothing.

Two adjacent gaps from the same audit:

- The raw browser leaves rejected `-v`, even though `cdp.ts:121,357` already gate diagnostics on `isVerbose()`. Reaching them required setting `WEBCMD_VERBOSE` by hand.
- `auth status` / `auth refresh` run browser and daemon probes but rejected `-v`.

## Change

- **`enableVerbose()` in `logger.ts`** — one place applying the `WEBCMD_VERBOSE` contract that `isVerbose()` already reads. Passing `false` leaves the environment untouched, so an exported `WEBCMD_VERBOSE=1` is not cancelled by a flagless invocation. Child processes inherit it.
- **Hosted** — apply it in `dispatchHosted`, and add request/response/failure diagnostics to `HostedClient.authenticatedFetch` so the flag is actually observable:
  ```
  [verbose] hosted → POST /v1/execute
  [verbose] hosted ← POST /v1/execute 200 (128ms)
  ```
  These carry method, path, status and elapsed time only — never the bearer token, the workspace header, or request/response bodies. A test pins that.
- **Browser** — `-v` on the bridge/CDP leaves (`tabs`, `bind`, `run`, `snapshot`, `close`). Filesystem-only leaves (`init`, `fork`) deliberately keep no flag rather than advertising a no-op, which is the thing #174 asks to avoid.
- **Auth** — `-v` on `auth status` and `auth refresh`, which drive that same browser/daemon stack.

`-v` stays a local concern: it is **not** added to the `/v1/execute` request body, and a test pins that too, so a local debugging flag cannot drift into the server contract.

## Note on the browser catalog

The browser flag is declared in both `cli.ts` and `browserCommandCatalog`. Local help comes from Commander while hosted help is rendered from the catalog, and `runner.test.ts` asserts the two are byte-identical — declaring it in one place only fails that parity test. `browserOptionFlags` grows a `verbose` case so hosted renders the `-v` short form, and the catalog entry carries an explicit `false` default to reproduce local's `(default: false)` suffix.

## Acceptance criteria

- [x] Hosted `-v` is effective or explicitly unsupported — now effective, with observable diagnostics.
- [x] Browser `-v` enables existing bridge/CDP diagnostics.
- [x] Any auth `-v` support has observable diagnostics — both probes go through the browser/daemon stack.
- [x] No command advertises a no-op verbose flag — hosted was the no-op; `init`/`fork`/`verify` intentionally stay without one.
- [x] Local and hosted behavior is covered by tests.

## Verification

- `npm run typecheck` clean.
- `npm test`: 5636 passing, +23 from this change. The 21 remaining failures are pre-existing on this Windows checkout (`EPERM` symlink and related environment limits) — identical file-for-file and count-for-count with a stashed clean tree at the same commit, so none are introduced here.
- All CI gate scripts pass: `check-community-plugins`, `check:codex-plugin`, `check:plugin-parity`, `check:hosted-contract`, `check:package-bin`, `check:silent-column-drop`, `check:typed-error-lint`.
- Manually confirmed `browser tabs --help`, `browser snapshot --help`, and `auth status --help` now list `-v, --verbose`.

## Scope

Sub-issue of #171. This is the functional half; #173 covers the skill/help documentation and explicitly says to update guidance "as the functionality sub-issues land", so it is left out of this PR.


ankitranjan7:
Thank you @Agnik47 — \`-v\` being a no-op in hosted/browser/auth was a real agent trap.

Merged \`main\` in cleanly (no conflicts). Landing this now.

PR #316, opened 2026-08-14
fix(local-cloak): honour waitUntil when opening a tab with a url

### LINKED ISSUE #210: ``waitUntil: 'load'is still hardcoded innewPage— the #106 fix reachednavigateonly, sotab new --url still hangs on never-idle sites

Agnik47:
Fixes #210

## The bug

#106 reported `navigate` hard-coding `waitUntil: 'load'`. That fix landed, but the same root cause had a second call site that was never covered:

```ts
// session-manager.ts — navigate, fixed by #106/#107
await lease.page.goto(url, { waitUntil });          // threaded through

// session-manager.ts — newPage, still hardcoded
await acquired.page.goto(input.url, { waitUntil: 'load' });
```

The gap was structural rather than a missed edit — `newPage` had no `waitUntil` in its signature at all, and its only caller dropped the field even though it was sitting on the command object being read:

```ts
// actions.ts — 'tabs' / 'new'
const lease = await manager.newPage({
  ...
  url: command.url,
  // command.waitUntil never passed
});
```

So `webcmd browser navigate --wait-until none` works while `webcmd browser tab new --url <url>` still blocks until the load event, and opening a tab on a site that never goes idle hangs — the exact failure #106 was filed about.

## The fix

`newPage` / `newPageAttempt` now take a `NewPageInput` carrying an optional `waitUntil`, defaulting to `'load'` so every existing caller is unaffected, and `actions.ts` passes the command's value through.

I also folded in the related inconsistency the issue flagged. The command-vocabulary (`'load' | 'none'`) → Playwright-vocabulary (`'load' | 'commit'`) translation existed only as an inline ternary at the navigate call site, which is part of why the second site was easy to miss. Both call sites now go through one `toGotoWaitUntil` helper, so a future `waitUntil` value reaches every Playwright-backed navigation at once.

One deliberate scoping decision: I kept the helper at the local-cloak boundary rather than unifying all three files the issue mentions. `cdp.ts` and `page.ts` implement webcmd's own `BasePage`, whose `goto` genuinely takes `'load' | 'none'` — they aren't driving Playwright, so their vocabulary is correct for their layer and collapsing it would be wrong. The two vocabularies only need reconciling where local-cloak hands off to a real Playwright page, and that is now a single function. The session manager stays in Playwright vocabulary, consistent with `navigatePage`.

## Verification

- 4 new tests in `session-manager.test.ts`, driven end-to-end through `dispatchCloakAction` so they cover the `actions.ts` plumbing that was actually broken, not just the session-manager signature.
- Against unfixed code, exactly one fails — `maps waitUntil none to commit when opening a tab with a url`. The other three (tab default, navigate `none`, navigate default) pass both before and after, which is the point: they are regression guards proving the default path and the #106/#107 navigate fix are unchanged.
- `npm run typecheck` clean; `check:typed-error-lint` reports `new=0`.
- Full `--project unit` suite: **2477 passed** with this branch vs **2473** on `main` — exactly the 4 added tests. The 20 pre-existing failures across 6 unrelated files (`discovery`, `docs-sync-review-cli`, `hosted/files`, `hosted/runner`, `plugin`, `site-memory/local-store`) are identical on a clean `main` checkout and are untouched by this change.


ankitranjan7:
Thank you @Agnik47 — this closes the leftover `waitUntil: 'load'` hole on `tab new --url` after #106/#107.

Merged `main` into the branch. Landing this now.

PR #321, opened 2026-08-15
fix(site-memory): lock site memory writes across processes

### LINKED ISSUE #320: [Bug]: Silent data loss in site memory due to cross-process race condition
# Description

Site memory writes are protected by an in-memory promise chain, not by anything the filesystem can see. Two `webcmd` processes writing to the same site therefore overwrite each other's work, and the losing write disappears with no error and exit code 0.

`withPathLock()` is the guard, and it is a plain `Map` held in one process's heap:

```ts
// src/site-memory/local-store.ts:60
const writeChains = new Map<string, Promise<void>>();

// src/site-memory/local-store.ts:183
async function withPathLock<T>(target: string, fn: () => Promise<T>): Promise<T> {
  const previous = writeChains.get(target) ?? Promise.resolve();
  const next = previous.catch(() => undefined).then(fn);
  const settled = next.then(() => undefined, () => undefined);
  writeChains.set(target, settled);
  settled.then(() => {
    if (writeChains.get(target) === settled) writeChains.delete(target);
  });
  return next;
}
```

A second `webcmd` process starts with its own empty `writeChains`. It has no way to learn that another process is mid-write, so the two proceed in parallel.

That would be harmless if the writes were blind overwrites. They are not — `updateText()` and `updateJson()` are read-modify-write:

```ts
// src/site-memory/local-store.ts:148
async function updateText(
  site: string,
  path: 'notes.md',
  opts: LocalStoreOptions,
  update: (existing: string) => string,
): Promise<void> {
  const root = await ensureSiteRoot(site, opts);
  const target = join(root, path);
  return withPathLock(target, async () => {
    await atomicWrite(target, update(await readText(target)));
  });
}
```

That inner line is a read (`readText`), a modify (`update`), and a write (`atomicWrite`) — three steps that must not be split by another writer. `updateJson()` at `:161` has the same shape.

So two processes interleave like this:

```
process A: read notes.md   -> "" (or existing content)
process B: read notes.md   -> same content
process A: write content + note-A
process B: write content + note-B      <- built from the stale read; note-A is gone
```

`atomicWrite()` (`src/site-memory/local-store.ts:194`) is written carefully — it writes to a temp file named with `process.pid` + `randomUUID()` and then `rename()`s it, so the file is never torn or half-written. That is correct as far as it goes, and it is probably why this looked safe. But `rename()` gives atomicity of a *single* write, not isolation across a read-modify-write pair. The result is last-writer-wins, and last-writer-wins on a read-modify-write means **lost updates**.

## Reproduction

Twenty concurrent `webcmd site note add` calls against one site. Every process exits 0; only some of the notes survive.

```bash
#!/usr/bin/env bash
# Run from a webcmd checkout after `npm install && npm run build`
export HOME=$(mktemp -d)
WEBCMD="node $PWD/dist/src/main.js"

for i in $(seq 1 20); do
  $WEBCMD site note add example.com --text "note-$i" &
done
wait

echo "notes written  : 20"
echo "notes surviving: $(grep -c '^note-' "$HOME/.webcmd/sites/example.com/notes.md")"
```

Output across three consecutive runs:

```
notes written  : 20
notes surviving: 18
---
notes written  : 20
notes surviving: 19
---
notes written  : 20
notes surviving: 18
```

It is nondeterministic — it depends on how the reads and writes interleave — so it does occasionally pass. Surviving notes across ten runs: **20, 19, 19, 19, 18, 18, 18, 17, 17, 13** out of 20. Nine of the ten lost data, the worst run losing 35% of it. No run printed a warning, and no process exited non-zero.

The same failure hits `endpoints.json` through the `updateJson()` path:

```bash
for i in $(seq 1 20); do
  $WEBCMD site endpoint set example.com "ep$i" --url "https://example.com/api/$i" --method GET &
done
wait
# keys in endpoints.json across three runs: 17, 19, 19  (expected 20)
```

## Affected files

| Location | What is affected | Why |
|---|---|---|
| `src/site-memory/local-store.ts:60`, `:183` | `writeChains`, `withPathLock()` | The lock is a per-process in-memory `Map`. Nothing coordinates across processes — no lockfile, no `O_EXCL`, no `flock`. |
| `src/site-memory/local-store.ts:148`, `:161` | `updateText()`, `updateJson()` | The read-modify-write functions. These are the only two places where a stale read can silently erase a concurrent write. |
| `src/site-memory/local-store.ts:63`, `:70`, `:87`, `:96` | `appendNote()`, `setEndpoint()`, `markEndpointStale()`, `addFieldMapping()` | The four public writers routed through `updateText`/`updateJson`. All four can lose data. |
| `src/site-memory/local-store.test.ts:138` | `"survives two concurrent appendNote calls"` | Covers only the in-process case, which already works. See below. |

Not affected, for scope clarity:

- `writeSiteFile()` (`src/site-memory/local-store.ts:175`), used by `putVerifyFixture()` and `addResponseSample()`, writes a complete body rather than modifying an existing one. Last-writer-wins there is defensible, and `addResponseSample()` generates a unique path per call anyway.
- Hosted mode is unaffected. `src/hosted/runner.ts:495` registers a server-side backend; this is specific to the local backend registered at `src/cli.ts:624`.

## Why the existing test does not catch it

```ts
// src/site-memory/local-store.test.ts:138
it('survives two concurrent appendNote calls', async () => {
  const homeDir = await tempHome();
  await Promise.all([
    appendNote({ ...base, homeDir, text: 'alpha' }),
    appendNote({ ...base, homeDir, text: 'beta' }),
  ]);

  const body = await readNotes(homeDir);
  expect(body).toContain('alpha');
  expect(body).toContain('beta');
});
```

Both calls run inside a **single** process via `Promise.all`, which is exactly the case `writeChains` handles correctly. The test passes, and its name reads as though concurrent writes are covered, while the cross-process case is untested and broken. Worth noting mainly because it explains how this got through review.

## Why a fix is necessary

This is silent data loss in the feature whose entire purpose is to accumulate knowledge over time. Site memory is what an agent has learned about a site — verified endpoints, field meanings, notes — and losing it is not a cosmetic failure.

It does not require an exotic setup. Any of these hit it today:

- Two agent sessions in two terminals working on the same site.
- CI or scripted runs invoking `webcmd` tasks in parallel.
- One agent running several `webcmd` calls concurrently rather than sequentially.

Because there is no error and no non-zero exit, nothing surfaces the loss. A user only notices later, when an endpoint they recorded is missing, and at that point there is no signal pointing at concurrency as the cause.

## Suggested fix

The data loss can be closed entirely inside `local-store.ts`, with no dependency and no change to the on-disk format:

1. Keep `writeChains` as the in-process fast path — it works and it avoids lock churn for the common single-process case.
2. Wrap the read-modify-write in `updateText()` and `updateJson()` in a cross-process lock: create `<target>.lock` with `fs.open(..., 'wx')` (`O_EXCL`, atomic create-if-absent), retry with backoff when it exists, and break locks older than a timeout by `mtime` so a crashed process cannot wedge site memory permanently. Release in a `finally`.
3. Critically, move the **read** inside the lock, so the sequence becomes acquire → read → modify → write → release. Locking only the write would not fix anything.

Sketch:

```ts
async function withCrossProcessLock<T>(target: string, fn: () => Promise<T>): Promise<T> {
  const lockPath = `${target}.lock`;
  // acquire: open(lockPath, 'wx') with retry/backoff; if EEXIST and the lock's
  // mtime is older than STALE_MS, unlink and retry once.
  try {
    return await fn();
  } finally {
    await unlink(lockPath).catch(() => undefined);
  }
}
```

`withPathLock()` then composes the two, so callers do not change:

```ts
return withPathLock(target, () => withCrossProcessLock(target, fn));
```

If a dependency is acceptable, `proper-lockfile` handles the stale-lock and retry logic and would be a smaller diff.

A test for the real case would need to spawn two child processes rather than use `Promise.all` — the reproduction script above is the shape of it, and it fails before the fix and passes after.

I am happy to open a PR for this if the approach sounds right.

## Environment

- webcmd `0.7.1` (`52da25b`)
- Node.js `v24.14.0`
- Linux (Fedora, kernel 7.1.8)
- Local site-memory backend (default)

Kaushik2003:
Fixes #320 

## The problem

I found that site memory writes are guarded by `writeChains`, a `Map` that only exists in one process's memory. When a second `webcmd` process starts, it gets an empty `Map`, so it has no way of knowing that another process is in the middle of a write.

This matters because `updateText()` and `updateJson()` are read-modify-write operations. Each call reads the file, changes it, and writes it back:

```
process A: read notes.md   -> current content
process B: read notes.md   -> same content
process A: write content + note-A
process B: write content + note-B     <- built from the stale read, note-A is gone
```

`atomicWrite()` renames a temp file into place, so a file is never left half-written. But that only makes a single write atomic - it doesn't stop another process from reading in between. The result is a lost update, with no warning and exit code 0.

## The fix

I added a lock that the filesystem itself can see, and moved the read inside it.

1. Before a read-modify-write, I create `<file>.lock` using `open(path, 'wx')`. That flag only creates the file if it doesn't already exist, and the OS guarantees only one process can win.
2. If the lock already exists, I wait a few milliseconds and try again, up to 15 seconds.
3. Read, modify, and write while holding the lock, then delete the lock in a `finally`.
4. If a lock is left behind by a crashed process, I break it and move on. I detect that in two ways:
   - The lock file records the owner's pid and hostname. If the owner is on this machine and no longer running, the lock is broken on the next try.
   - Any lock older than 10 seconds is broken regardless. The work inside the lock is one small read plus one rename, so it takes milliseconds. Ten seconds means the owner is gone.

I kept the in-process `writeChains` and run it first. It's free, and it settles same-process callers without touching the disk. The file lock only kicks in when a write actually needs to be coordinated across processes.

I didn't add a new dependency. `proper-lockfile` would have worked, but this came out to about 170 lines, and I'd rather keep the dependency list small.

## Files I touched

**`src/site-memory/file-lock.ts` (new)**

This is the lock itself: `withFileLock(target, fn)` acquires, runs `fn`, and releases in a `finally`. It contains the retry loop, the stale-lock breaking, and the busy error.

Two details worth a reviewer's attention:

- *Release checks ownership.* The lock file holds a random token. On release, a process only deletes the lock if the token still matches its own. Without this check, a process whose lock was broken would end up deleting its successor's lock on the way out, and writes would stop being serialized again.
- *Breaking a lock re-checks mtime.* I read the mtime once when deciding a lock is stale, and again right before `unlink`. If it changed in between, someone released and re-took the lock, so I leave it alone.

**`src/site-memory/local-store.ts`**

- Added `withWriteLock()`, which composes the two locks: `withPathLock(target, () => withFileLock(target, fn))`.
- `updateText()` and `updateJson()` now use `withWriteLock` instead of `withPathLock`. This is the actual fix - the read moves inside the lock, because locking only the write wouldn't change anything.
- I left `writeSiteFile()` on `withPathLock` alone. It writes a full body rather than modifying an existing one, so there's no stale read to lose. Its callers (`putVerifyFixture`, `addResponseSample`) are unaffected.
- Renamed `isTempWritePath()` to `isInternalWritePath()` and had it also skip `*.lock`. Readers walk the site directory, so without this a lock file would show up as a site memory entry in `webcmd site memory show` and `list` while a write is in flight, or after a crash. This is a behavior change: lock files are now invisible to readers, the same way `.tmp` files already are.

**`src/site-memory/__fixtures__/concurrent-writer.mts` (new)**

A small script the tests run in child processes. It takes a home directory, a site, a mode (`note` or `endpoint`), a label, and a count, then writes that many entries. I made it a `.mts` file so `tsconfig`'s `src/**/*.ts` include doesn't pull it into the build.

**`src/site-memory/file-lock.test.ts` (new)**

Six tests for the lock on its own: overlapping sections run one at a time, the lock is released when the inner function throws, a lock owned by a dead process is broken, a lock older than the stale window is broken, a live holder produces `SITE_MEMORY_BUSY` and writes nothing, and a lock taken over by someone else isn't deleted on release.

## Why I changed `src/site-memory/local-store.test.ts`

This file needed changing because its existing concurrency test could never have caught this bug, even though its name suggested it did.

The old test was called `survives two concurrent appendNote calls`, and it did this:

```ts
await Promise.all([
  appendNote({ ...base, homeDir, text: 'alpha' }),
  appendNote({ ...base, homeDir, text: 'beta' }),
]);
```

Both calls run in **one** process, which is the case `writeChains` already handled correctly. The test passed before this PR and still passes after it. Reading the test list, it looked like concurrent writes were covered, when in fact the cross-process case was untested and broken.

Here's what I changed:

1. **Renamed** it to `survives two concurrent in-process appendNote calls`. Same test, honest name. It still has value, it just covers a different case.
2. **Added `keeps every note when separate processes append at the same time`.** This spawns 6 real child processes (via `tsx` and the fixture script above), each writing 4 notes, then checks all 24 notes are present. Each child writes several times on purpose, so the processes overlap during the writes and not only at startup.
3. **Added `keeps every endpoint when separate processes write at the same time`.** Same shape, through the `updateJson()` path, checking all 24 keys are in `endpoints.json`.
4. **Added `hides write lock markers from readers`.** This drops a `notes.md.lock` file into a site directory and checks that `show` and `list` ignore it, mirroring the existing `.tmp` test.
5. **Added a small helper** (`runWriters` / `expectedWrites`) at the bottom of the file to spawn the children.

The two new cross-process tests are the ones that matter. They fail on `main` and pass on this branch.

## Tests I ran

**The new tests, against the old code.** I reverted the two `withWriteLock` lines back to `withPathLock` and ran the suite. Both cross-process tests failed. I restored the fix and they passed. This confirms the tests are actually testing something real.

**The reproduction from the issue, against a real build.** 20 concurrent processes per run:

```
notes,     3 runs: 20/20, 20/20, 20/20    (was 18, 19, 18)
endpoints, 3 runs: 20/20, 20/20, 20/20    (was 17, 19, 19)
```

No lock files were left behind in any run.

**Full suite.** `npm test` - 443 files, 5653 passed, 2 skipped. `npm run typecheck` clean. `node scripts/check-typed-error-lint.mjs` clean, no new entries against the baseline.

**Flakiness.** I ran `src/site-memory` five times in a row, and got 27 tests passing each time. The two cross-process tests take about 200ms each.

## One decision worth flagging

When the lock can't be acquired, I made it fail loudly instead of writing anyway - it throws `SITE_MEMORY_BUSY` (exit code 75), same pattern as `SESSION_BUSY`. Writing anyway would bring back the silent data loss this PR is meant to fix.

This shouldn't trigger in normal use - a lock only turns into an error after 15 seconds of real contention, and anything abandoned gets cleared after 10.

If you think this should be handled differently, let me know - happy to change it.


ankitranjan7:
Thank you @Kaushik2003 — this is a real silent-data-loss bug, and the lock is done carefully: \`wx\` acquire, read inside the lock, token-checked release, stale/dead-pid break, and an actual two-process test.

Merged \`main\` in cleanly. Landing this now.

PR #323, opened 2026-08-16
fix(browser): evict dead Profile runtime after a closed-target run failure

### LINKED ISSUE #314: browser run: recover when an apparently open page belongs to a dead context
## Summary

In Webcmd 0.7.1, raw `browser run` can retain and repeatedly reuse a Playwright page for which `page.isClosed()` is `false`, even though the underlying browser context/CDP connection is already closed.

The Session is created successfully and `browser run` can execute non-browser JavaScript, but every navigation fails with:

```text
QuickJS promise rejected: Target page, context or browser has been closed
```

A new Session on the same Profile later works, so this is a transient local runtime/session lifecycle failure rather than a destination-site failure.

This was observed in a Pi benchmark run and caused the controller to exhaust its browser attempts without acquiring any page evidence.

## Environment

- Webcmd: `0.7.1` (`/opt/homebrew/bin/webcmd`)
- Controller: `@earendil-works/pi-coding-agent 0.80.6`
- OS: macOS
- Profile: a long-lived shared `benchmark` Profile
- Explicit opaque Session per benchmark task

## Observed sequence

Using an explicit Session:

```bash
webcmd --profile benchmark --session <session-id> browser run --stdin <<'JS'
await page.goto('https://stackexchange.com/sites');
return { url: page.url(), title: await page.title() };
JS
```

failed repeatedly with the closed-target error. `browser tabs` returned `[]`.

A diagnostic run against the same leased object then returned:

```json
{
  "ok": true,
  "result": {
    "pageClosed": false,
    "contextPages": 1
  },
  "page": {
    "url": "about:blank"
  },
  "warnings": [
    {
      "code": "BROWSER_RUN_SNAPSHOT_FAILED",
      "message": "browserContext.newCDPSession: Target page, context or browser has been closed"
    }
  ]
}
```

Every subsequent `page.goto(...)` in that Session failed in the same way.

After the failed benchmark completed and cleaned up, creating a fresh explicit Session on the same `benchmark` Profile and navigating to `https://example.com` succeeded. This confirms the original failure was transient runtime state, not general connectivity.

## Likely recovery gap

`CloakSessionManager.getPage()` treats an existing entry as healthy when `page.isClosed()` is false. Raw `browser run` then executes directly against that lease. A disconnected/dead context can therefore masquerade as an open page.

There is closed-context retry logic around page creation and `navigatePage`, but a closed-context failure thrown inside `runBrowserProgram` does not appear to invalidate the Profile runtime or retry with a fresh page/context. The failed snapshot is downgraded to a warning, so even the CDP failure does not repair the runtime before the next command.

This differs from #293, which covers initial adapter navigation, and is the raw `browser run` path. It is related to the lifecycle class previously discussed in #276.

## Expected behavior

When `browser run` observes the closed-target signature from Playwright or snapshot CDP setup, Webcmd should:

1. invalidate the dead Profile runtime;
2. create a healthy replacement context/page;
3. retry safely when no user-visible side effects occurred, or return a structured reconnect error when replay is unsafe;
4. ensure the next command in the same Session does not receive the same dead lease.

At minimum, a failed run should evict the dead runtime so the next command can recover rather than repeatedly leasing the same invalid page.

## Suggested regression coverage

- Existing page reports `isClosed() === false`.
- `page.goto()` inside `browser run` rejects with the closed-target signature.
- `context.newCDPSession(page)` also rejects with the same signature.
- The Profile runtime is invalidated.
- The next raw browser command gets a replacement context/page and can navigate.
- Recovery is bounded and does not blindly replay a program after possible side effects.

rohan911438:
**Description:**

## Summary

Fixes #314.

Raw `browser run` leased pages directly and ran the program with no
recovery path: `CloakSessionManager.getPage()` treats a lease as
healthy whenever `page.isClosed()` is false, but a disconnected
context/CDP session can still report that. `dispatchCloakAction`'s
`run` case called `runBrowserProgram` with no try/catch, so a
closed-target failure never invalidated the Profile runtime the way
`navigatePage`/`newPage` already do — every subsequent raw browser
command on that Session kept leasing the same dead page.

## Changes

- Export `isClosedContextError` from `session-manager.ts`.
- Add `CloakSessionManager.evictDeadRuntime(profileId, context)` —
  invalidates the Profile runtime only if it still matches the caller's
  context (no-ops if it was already replaced, so it can't evict a
  healthy runtime that raced ahead of the failed lease).
- Wrap `runBrowserProgram` in `actions.ts`'s `run` case in try/catch:
  on the closed-target signature, evict the runtime, then rethrow.

We don't retry the program itself — it may have already caused side
effects, so blind replay isn't safe. The caller sees the original
error; the *next* command gets a fresh context/page instead of
repeating the same failure forever.

## Test plan

- [x] New `run-recovery.test.ts`: closed-target run failure evicts the
      runtime and the next `run` command gets a fresh page; an
      unrelated failure does not evict.
- [x] `npx tsc --noEmit` clean.
- [x] Existing `session-manager.test.ts` / `browser-run.test.ts` still pass.

### DIFF
--- a/src/browser/runtime/local-cloak/actions.ts
+++ b/src/browser/runtime/local-cloak/actions.ts
@@ -10,7 +10,7 @@ import {
 import { redactText, redactUrl } from '../../../observation/redaction.js';
 import { articleHtmlToMarkdown } from '../../../download/article-download.js';
 import { waitForDownload } from './downloads.js';
-import type { CloakSessionManager } from './session-manager.js';
+import { isClosedContextError, type CloakSessionManager } from './session-manager.js';
 import type { BrowserContext, Frame, Page as PlaywrightPage } from 'playwright-core';
 import { runBrowserProgram } from '../../run/runner.js';
 import { BROWSER_RUN_MAX_SOURCE_BYTES } from '../../run/types.js';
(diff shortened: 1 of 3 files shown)


ankitranjan7:
Thank you @rohan911438 — you nailed the diagnosis: \`page.isClosed()\` can lie, and a failed raw \`browser run\` never evicted the Profile runtime.

We landed #324 for the same bug because it also recovers *before* reuse (liveness probe in \`getPage\`) and when a successful run's snapshot sees a dead context. Your \"don't blindly replay a run that may have had side effects\" rule is the one we kept.

Closing this as superseded by #324. The work was useful.

PR #326, opened 2026-08-16
feat(cli): add -f/--format to webcmd profile list

rohan911438:
## Summary

Partial slice of #175.

Local `profile list` rejected `-f` entirely and only printed prose,
while hosted `profile list` already supports `-f json|yaml|csv|md`
(`src/hosted/runner.ts`) — the same command name behaving differently
between modes with no indication why. #175 explicitly calls out
"align local profile list with hosted profile-list formatting."

## Changes

- Add `-f, --format` (default `table`, prose output unchanged).
- Other formats render one row per profile through the shared output
  path. Row shape aligns with hosted's where the concept overlaps
  (`default`), plus locally-meaningful fields hosted doesn't have
  (`connected`, `runtimeVersion`) since Cloak profiles are live
  runtime connections, not persisted Cloud records.
- Disconnected saved aliases (already shown in the prose output) are
  included as `connected: false` rows so the structured view doesn't
  silently drop information the text view shows.
- The daemon-not-running/stale/no-profiles branches render `[]` for
  non-table formats rather than swallowing output, mirroring the
  existing `adapter status` empty-list convention.

## Scope note

Same slice-of-#175 approach as the `validate` and `daemon status` PRs.

## Test plan

- [x] New cases in `src/cli.test.ts`: `-f json` with no profiles
      (→ `[]`), and with a connected profile + a disconnected saved
      alias + a disconnected default (→ structured rows).
- [x] `npx tsc --noEmit` clean.
- [x] `npx vitest run --project unit src/cli.test.ts` — 104/104 pass.
- [x] Manual smoke: `webcmd profile list -f json` / `webcmd profile list`.

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -17,6 +17,7 @@ import {
 } from './command-presentation.js';
 import { parseOutputFormat } from './command-surface.js';
 import { render as renderOutput } from './output.js';
+import { saveProfileConfig } from './browser/profile.js';
 import * as pluginModule from './plugin.js';
 import * as discoveryModule from './discovery.js';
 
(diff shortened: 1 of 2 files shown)


adikulkarni006 (approved):

ankitranjan7:
Thank you @rohan911438 — `profile list -f` already landed on `main` via #306. Closing this companion slice as superseded.

ankitranjan7:
Superseded by #306.

PR #405, opened 2026-08-21
skills: SESSION_BUSY is two CLI invocations, not an in-page lock

### LINKED ISSUE #385: skills(browser): overlapping commands on one session must use SESSION_BUSY, not an in-page lock
## Why this issue exists

This came from the `local:session-concurrency` eval. The eval is trying to prove two separate things:

1. Two different sessions can run browser work in parallel.
2. Two overlapping CLI invocations against the same session produce structured `SESSION_BUSY`, followed by a safe retry.

Source reports:

- `/Users/ankitranjan/Work/agentrhq/evals/results/run-20260820T085734Z/report.md`
- `/Users/ankitranjan/Work/agentrhq/evals/results/run-20260820T103952Z/report.md`

## Current Webcmd state

- `SESSION_BUSY` exists as a structured CLI error.
- The error message says the session is busy because another holder is already driving it.
- The hint says: wait, then run the same command again.
- If the holder pid is live, the hint says not to force-close; if the holder is not live, `webcmd session close --force <session-id>` is presented as a last resort.
- The local runtime queues browser work by profile/session/site before execution, so session occupancy is a Webcmd runtime concern.
- Current skill/help text names `SESSION_BUSY`, but it does not make the two-CLI-invocation shape impossible to miss.

## Observed agent behavior

The successful run eventually did the right thing:

- Created two explicit sessions.
- Backgrounded two `webcmd --session <id> browser run` commands and proved the 3s waits overlapped.
- Started one long command on the Alpha session.
- Issued a second `webcmd --session <same-alpha-id> browser run` while the first was still holding the session.
- Received `SESSION_BUSY` naming the holder pid.
- Waited, retried the same second command without `--force`, and verified final count `2`.

The failed run shows the agent confusion:

- It created sessions, but did not create CLI-level contention.
- It tried to overlap work inside one `browser run` with two `page.evaluate(applyUpdate)` calls.
- Both in-page calls completed, so no `SESSION_BUSY` envelope appeared.
- The agent concluded the structured busy response did not exist in page code.
- The final count became `3`, because the retry happened after in-page writes rather than after a CLI busy/retry sequence.

## What went wrong

The agent modeled session occupancy as a page-level lock. It looked for busy state inside the browser page instead of in the Webcmd session runtime.

`SESSION_BUSY` is not produced by two promises inside one `browser run`. It is produced when a second Webcmd CLI invocation tries to drive the same session while another invocation still owns it.

The recurring failure pattern is:

1. The agent sees a concurrency task.
2. It creates two sessions correctly.
3. For same-session contention, it stays inside one browser program.
4. It invents an in-page lock or in-page overlap.
5. It never asks the CLI to arbitrate the session.

## Why this matters

This is a product affordance issue more than an eval-specific skill issue. Agents should not need deep skill text to understand where the lock lives. The CLI/runtime should make it hard to miss that session contention is between Webcmd invocations, not page scripts.

Related product concern: `SESSION_BUSY` must be returned before browser work starts when a session is already held.

## Solution directions

The preferred solution should put the durable behavior in the CLI/runtime first.

Possible directions:

1. Product/runtime: ensure every mutating browser command checks session ownership before doing browser work and returns structured `SESSION_BUSY` with holder details.
2. CLI help/error text: make the error say that the caller should wait for the listed holder, then retry the same command. `--force` is only for a dead holder.
3. `session list`: make active holders visible enough that an agent can inspect the session state without inventing page locks.
4. Minimal skill text, if needed: one troubleshooting line saying `SESSION_BUSY` comes from a second `webcmd --session <id> ...` process, not from `page.evaluate`.

- Is the runtime/error surface already strong enough after the product fix, making the skill PR unnecessary?
- Should there be a CLI-visible way to show the current holder command/pid in `session list`?
- Should the docs include a two-terminal recipe, or is that too much eval-specific instruction?

## Acceptance criteria

- A second CLI invocation on a held session returns structured `SESSION_BUSY` before browser work starts.
- The error tells the agent to wait for the holder and retry the same command.
- The agent does not implement an in-page mutex or use `page.evaluate` as the safety signal.
- The safe retry path completes without `--force` and without an extra write.

## Non-goals

- No fixture-specific wording.
- No `data:` localStorage policy changes here.
- No encouragement to force-close a live holder.

ayushsingh82:
Fixes #385.

## Problem

On a same-session concurrency task, an agent modeled Session occupancy
as a page-level lock: it stayed inside one `browser run` and overlapped
two `page.evaluate` calls instead of issuing a second
`webcmd --session <id> ...` CLI invocation. Both in-page calls
completed, no `SESSION_BUSY` envelope appeared, and the agent concluded
the structured busy response didn't exist for this case — writing
twice instead of retrying safely.

`SESSION_BUSY` is produced by the Webcmd runtime arbitrating a second
CLI invocation against a Session another invocation still holds. It
has no relationship to anything that happens inside one `run`'s
`page.evaluate` calls — the runtime already returns it correctly today
(confirmed against `origin/main`); the gap is that skill text didn't
make the two-CLI-invocation shape impossible to miss.

## Fix

- `webcmd-browser`: added a worked two-terminal example (Terminal A
  holds the Session with a long `run`, Terminal B issues a second
  `--session` command against the same id), a `Do Not` line against
  implementing occupancy inside the page, and a sharper `SESSION_BUSY`
  troubleshooting row that says "retry the same command" and "do not
  lock inside `page.evaluate`."
- `webcmd-usage`: reworded the `SESSION_BUSY` explanation to name the
  two-invocation shape explicitly instead of "another holder."

Skill-doc only change; no runtime/CLI behavior touched.

## Test plan

- `npm run typecheck` — clean.
- `npm run test` — 447 files / 5757 tests pass.
- `make verify` — skill build (`skill-src/` → `skills/`) in sync.
- Updated `src/skills.test.ts` to assert the new in-page-lock anti-
  pattern text and the two-invocation wording.

### DIFF
--- a/skill-src/webcmd-browser/SKILL.src.md
+++ b/skill-src/webcmd-browser/SKILL.src.md
@@ -153,6 +153,20 @@ JS
 
 If you split work across calls, use fresh snapshots between page transitions. Do not assume an observation from a prior route is still valid.
 
+To contend for one Session, issue a second `webcmd --session <session-id> …` while the first invocation is still running. Do not start a second driver inside one `run` (`Promise.all` of two writes, a homemade busy flag, or `page.evaluate` locking).
+
+```bash
+# Terminal A — long run holding the Session
+webcmd --session session_abc browser run --file long.js
+
+# Terminal B — second command against the same Session
+webcmd --session session_abc browser run --stdin --no-snapshot-diff <<'JS'
+return { url: page.url() };
+JS
+```
+
+If the second command returns `SESSION_BUSY`, wait for the listed holder, then retry that same command. Do not `--force` close a live holder.
+
 ---
 
 ## Sitemaps
@@ -250,6 +264,7 @@ Use `run` and inspect `page.frames()`; target the frame by URL/name and keep ifr
 
 - **Do not submit forms via `page.evaluate(() => document.forms[0].submit())`.** Modern sites intercept real click/submit events and silently drop direct DOM submission. Use Playwright locators and verify the post-action state.
 - **Do not reuse observations across a page transition.** Navigations, form submits, SPA route changes, login, and human handoff invalidate earlier observations. Take a fresh snapshot.
+- **Do not implement Session occupancy inside the page.** Overlap two CLI invocations on the same `--session` id. A homemade busy flag is not `SESSION_BUSY`.
 - **Do not run a trigger before arming the waiter.** If a request matters, create `page.waitForResponse(...)` before the click/fill/keypress that triggers it.
 - **Do not trust autocomplete or masked inputs blindly.** Fill/type can appear to work while the app rejects the value. Verify visible text, `inputValue()`, or post-action state.
 - **Do not solve CAPTCHA or auth challenges programmatically.** Use human handoff and verification.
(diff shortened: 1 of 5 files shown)


ayushsingh82:
@ankitranjan7 mind taking a look? This closes #385.

PR #403, opened 2026-08-21
fix(cli): session not found names the owning profile

ayushsingh82:
Fixes #388.

## Problem

`SESSION_NOT_FOUND` said `Session not found: <id>` even when the id
existed under a different profile, and `session list` gave no visible
signal that its output was scoped to the selected profile. Agents that
created sessions under a named profile and later ran `session close`
or `session list` without repeating `--profile` read the error as
"session doesn't exist" and either retried the same wrong command or
inspected unrelated default-profile state.

## Fix

- `SessionNotFoundError` now looks up the session's actual owning
  profile in the store and, when it differs from the selected one,
  names it in the hint with the exact retry command:
  `webcmd --profile <owner> session close <id>`. Falls back to the
  existing generic hint when the id doesn't exist under any profile.
- `session list` table/JSON output now includes a `profileId` column,
  so the output doesn't read as global when it's profile-scoped.
- Updated `docs/troubleshooting.mdx` and the `webcmd-browser` skill's
  troubleshooting table with the new behavior.

## Test plan

- `npm run typecheck` — clean.
- `npm run test` — 447 files / 5757 tests pass.
- `make verify` — skill build (`skill-src/` → `skills/`) in sync.
- Added coverage in `src/browser/sessions.test.ts` for both the
  owning-profile hint and the fallback (unknown session id) hint.

PR #406, opened 2026-08-21
fix(cli): name the owning Profile when a Session lookup misses

### LINKED ISSUE #388: cli: session commands without --profile should name the owning profile
## Why this issue exists

This came from the `local:profile-separation` eval. The task required two separate browser profiles, separate sessions, restart persistence, and no cross-profile contamination.

Source reports:

- `/Users/ankitranjan/Work/agentrhq/evals/results/run-20260820T085734Z/report.md`
- `/Users/ankitranjan/Work/agentrhq/evals/results/run-20260820T103952Z/report.md`

## Current Webcmd state

- `webcmd session create` now requires the selected profile to exist; unknown profile creation paths are clearer than before.
- `webcmd session list` is scoped to the selected profile.
- Empty table output says `No browser Sessions found for Profile <profileId>`.
- The table columns for `session list` are still `id`, `kind`, `runtimeState`, and `handoff`; the table does not show `profileId`.
- `SESSION_NOT_FOUND` currently says `Session not found: <id>` and hints to run `webcmd --profile <selected-profile> session list`.
- Current `SESSION_NOT_FOUND` does not search the stored session registry for the owning profile and does not print the exact `--profile <owner>` retry command.

## Observed agent behavior

The browser isolation itself can work:

- Agents created separate profiles such as `personal-run-20260820t085734z` / `work-run-20260820t085734z` and `personal-run-20260820t103952z` / `work-run-20260820t103952z`.
- They drove browser work with `--profile <alias>`.
- Each profile retained its own `localStorage` value after closing and recreating sessions.
- Cross-profile storage stayed clean in the successful parts of the eval.

The confusing part was session/profile cleanup and inspection:

- In `run-20260820T085734Z`, `webcmd session close <id>` without `--profile` returned `Session not found` for sessions that belonged to non-default profiles.
- The agent repeated the same close attempts and got the same error.
- `webcmd session list -f json` without `--profile` dumped many idle sessions from `default`, not the sessions from the newly created profile aliases.
- In `run-20260820T103952Z`, unscoped `profile list` / `session list` showed unrelated older profiles and default-profile sessions. The agent described unrelated `085734z` profiles as already existing, causing a tenant-clean failure.

## What went wrong

The CLI made the selected profile too implicit.

When the agent had a valid session ID from one profile but ran `session close` or `session list` without that same `--profile`, the error looked like the session did not exist. It did not say that the session existed under another profile, and the list output did not make it obvious that it was only showing the selected/default profile.

The recurring pattern is:

1. Agent creates sessions under named profiles.
2. Later cleanup omits `--profile`.
3. CLI says `SESSION_NOT_FOUND` from the default/selected profile context.
4. Agent retries the same wrong command or inspects unrelated default sessions.
5. The task drifts into stale profile/session state.

## Why this matters

This is exactly the kind of thing the CLI should handle. A weak skill should not have to remember every session/profile pairing. If the CLI has the session registry, it can name the owning profile or make the selected-profile scope obvious.

This is also a safety issue for hosted or multi-tenant workflows: unclear profile scope encourages agents to inspect or describe unrelated state.

## Solution directions

The preferred solution is product-first.

Possible directions:

1. `session close <id>`: if the ID exists under a different profile, return structured `SESSION_NOT_FOUND` with a hint naming the owning profile and the exact `--profile <owner>` retry shape.
2. `session list`: include `profileId` in table/JSON output or otherwise state clearly that the list is scoped to the selected profile.
3. Empty list copy: say `No browser Sessions found for Profile <profileId>` so agents do not read it as global absence.
4. Discovery path: make `profile list` and `session list` help text explicit about profile scoping.
5. Minimal skill text, if needed: `SESSION_NOT_FOUND` after create usually means the follow-up command omitted the same `--profile` used on create.

- Should `session list` without `--profile` ever show all profiles, or should it stay scoped and make the scope impossible to miss?
- Should the owner-profile lookup happen only locally, or also through the daemon/provider path?
- Is the small browser-skill troubleshooting line still needed once the CLI error is explicit?

## Acceptance criteria

- Closing a session from the wrong profile names the owning profile when known.
- The retry hint includes `webcmd --profile <owner> session close <id>`.
- Session list output does not look global when it is profile-scoped.
- Agents stop retrying the same wrong close command and stop inspecting unrelated default sessions.

## Non-goals

- No `profile create` changes here.
- No broad profile cleanup command.
- No eval-specific profile aliases in user-facing text.

Agnik47:
Fixes #388.

## Problem

A Session ID only ever resolves inside the selected Profile, so a command that omits `--profile` reports `Session not found` for a Session that plainly exists. The eval runs in the issue show the loop this creates: the agent has a valid ID from `session create`, drops the flag on the follow-up `session close`, reads the error as "already gone", and then either retries the identical command or starts inspecting unrelated `default` Sessions.

`session list` fed the same confusion from the other side — the rows were always scoped to the selected Profile, but the table never said so, so a list of `default` Sessions read as the whole machine.

## Fix

Every Profile's Sessions live in one `browser-sessions.json` keyed by `profileId`, so when the scoped lookup misses, the owning Profile is already in hand — no daemon or provider round trip is needed to name it.

- `SessionNotFoundError` takes an optional owner. When known, the message becomes `Session not found in Profile <selected>: <id>` and the hint names the owner plus the exact retry, `webcmd --profile <owner> session close <id>`. With no owner the message and hint are byte-for-byte unchanged.
- Both throw sites are covered: `require`, and the private `requireMutable` behind `remove` / `touch` / handoff updates — so the close path cannot fall back to the anonymous message.
- The daemon provider (`local-cloak/provider.ts`) and the CLI local fallback both resolve through this same store, so one store-level change serves both paths. That also settles the issue's open question about local-only versus daemon lookup: it is local, and it is correct for both.
- `session list` gains a `profileId` table column. JSON and YAML already carried the field; `columns` only affects table/markdown/csv.
- `docs/troubleshooting.mdx` documents the new `SESSION_NOT_FOUND` behaviour.

## Acceptance criteria

- [x] Closing a Session from the wrong Profile names the owning Profile when known.
- [x] The retry hint includes `webcmd --profile <owner> session close <session-id>`.
- [x] Session list output does not look global when it is Profile-scoped.
- [x] Agents get a one-step retry instead of a dead end.

## Open questions from the issue

- **Should `session list` without `--profile` ever show all Profiles?** Left scoped, with the scope now visible in the table. Showing every Profile by default is the behaviour the issue calls a safety problem for hosted and multi-tenant use.
- **Local-only or daemon lookup?** Local. Both paths share the store, so it is already consistent.
- **Is the skill troubleshooting line still needed?** Not addressed here — the CLI error is now explicit, so that line can be revisited separately.

## Tests

Three store-level tests (owner named on `require`, owner named on the mutating path, generic hint preserved when no Profile owns the ID) and one CLI-level test that drives `session close` without `--profile` end to end.

155/155 pass across `sessions.test.ts`, `cli.test.ts`, and `session-docs-sync.test.ts`; `tsc --noEmit` is clean.

Note for reviewers on Windows: the four symlink tests in `src/site-memory/local-store.test.ts` fail on a clean `main` in my environment (symlink permissions), unrelated to this change.

## Non-goals

No `profile create` changes, no broad profile cleanup command, and no eval-specific aliases in user-facing text.

### DIFF
--- a/docs/troubleshooting.mdx
+++ b/docs/troubleshooting.mdx
@@ -48,7 +48,7 @@ Common Session codes:
 | `SESSION_REQUIRED` | A raw browser command needs a root Session selector. | Run `webcmd session create -f json`, then retry as `webcmd --session <session-id> browser ...`. |
 | `INVALID_SESSION_SELECTOR` | The selector is not an opaque Webcmd Session ID. | Use an ID returned by `webcmd session create` or `webcmd session list`. |
 | `SESSION_SELECTOR_POSITION` | `--session` was placed after the command name. | Move it before the command: `webcmd --session <session-id> browser ...`. |
-| `SESSION_NOT_FOUND` | The selected Session is missing for the current Profile. | Run `webcmd session list -f json`; create a new Session if needed. |
+| `SESSION_NOT_FOUND` | The selected Session is missing for the current Profile. | When the ID belongs to another Profile the hint names it — retry with `webcmd --profile <owner> session close <session-id>`. Otherwise run `webcmd session list -f json`; create a new Session if needed. |
 | `INVALID_SESSION_LIMIT` | `session list --limit` is outside 1-100. | Retry with a limit from 1 to 100. |
 | `SESSION_BUSY` | Another command is writing in the same Session or site scope. | Wait for the holder to finish; if it is dead, use `webcmd session close <session-id> --force` as the last resort. |
 | `SESSION_PAUSED_FOR_HUMAN_HANDOFF` | The Session is waiting for user action such as sign-in. | Finish the action in the browser, then run the returned verifier before retrying. |
--- a/src/browser/sessions.test.ts
+++ b/src/browser/sessions.test.ts
@@ -47,6 +47,44 @@ describe('LocalBrowserSessionStore', () => {
     expect(() => store.find('profile_work', 'work')).toThrowError(expect.objectContaining({ code: 'INVALID_SESSION_SELECTOR' }));
   });
 
+  it('names the owning profile when the session exists under a different one', () => {
+    // A cleanup command that omits `--profile` looks up a real Session ID in
+    // the wrong Profile. A bare "not found" sent agents into retrying the same
+    // command; naming the owner and the retry shape ends that loop.
+    const store = new LocalBrowserSessionStore({ baseDir: tempDir(), idFactory: () => 'session_a' });
+    const created = store.create('profile_work');
+
+    expect(() => store.require('profile_personal', created.id)).toThrowError(expect.objectContaining({
+      code: 'SESSION_NOT_FOUND',
+      ownerProfileId: 'profile_work',
+      message: `Session not found in Profile profile_personal: ${created.id}`,
+      hint: `Session ${created.id} belongs to Profile profile_work. Re-run the same command with \`--profile profile_work\`, for example \`webcmd --profile profile_work session close ${created.id}\`.`,
+    }));
+  });
+
+  it('names the owning profile on mutating lookups too', () => {
+    // `remove`/`touch`/handoff updates resolve the record through a separate
+    // lookup, so the close path must not fall back to the anonymous message.
+    const store = new LocalBrowserSessionStore({ baseDir: tempDir(), idFactory: () => 'session_a' });
+    const created = store.create('profile_work');
+
+    expect(() => store.remove('profile_personal', created.id)).toThrowError(expect.objectContaining({
+      code: 'SESSION_NOT_FOUND',
+      ownerProfileId: 'profile_work',
+    }));
+  });
+
+  it('keeps the generic hint when no profile owns the session', () => {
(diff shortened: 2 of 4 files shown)


ankitranjan7:
Rebased onto main and pushed. Changes made during the rebase:

- Kept main's fuller no-owner hint (from #420) as the fallback; the
  owner-named message and hint are unchanged from your version.
- Dropped the cli.test.ts case that asserted `session close` rejects.
  Since #420 that command is idempotent and resolves with
  {ok: true, alreadyClosed: true}, so the assertion could not hold. The
  three store-level cases in sessions.test.ts cover the owner naming,
  including the no-owner fallback.
- Updated the no-owner hint expectation in sessions.test.ts to match
  main's wording.

Verified locally: `npm run typecheck` clean, unit project 156 files /
2754 tests pass.

The `profileId` column you added to `session list` also closes the second
half of #388, so #403 is redundant — closing that one.

PR #415, opened 2026-08-22
feat(npm): add versions command

sreeramakhil:
## Description

Adds `webcmd npm versions <name>` to list published npm package versions newest-first with published date, `isLatest`, and npmjs URLs.

This PR was split before merge: the broken Reddit omnisearch source from the original branch has been removed, so the final diff is npm-only.

## Verification

- `npx vitest run plugins/npm/test/npm.test.js` -> 16 passed
- `npx vitest run plugins/npm/test/npm.test.js plugins/omnisearch/test/research.test.js` -> 17 passed
- `npm run typecheck` -> passed
- `npm run build` -> passed
- `npm run check:plugin-parity` -> passed
- `npm run check-community-plugins` -> passed
- `npm test` -> 448 files passed; 5,847 passed; 1 skipped
- `npm run check:package-bin` -> passed


ankitranjan7:
Split request. The npm `versions` command is good and I'd like to merge it;
the Reddit source doesn't work and shouldn't land.

npm versions — verified against the live registry, correct ordering and
isLatest, and not redundant with `npm package`. Two notes before merge:

1. Please drop the `request = fetch` parameter from `npmFetch` and
   `versionsNpm`. The tests in this PR already use `vi.stubGlobal('fetch')`
   for omnisearch — same approach works here without changing a production
   signature that no other npm command has.
2. On `react`, the default limit returns only canary/experimental builds
   with isLatest: false everywhere. Could you filter prereleases by default
   or add a `--stable` flag?

Reddit source — reddit.com/search.json returns 403 now. I ran the exact
fetch from this PR: 403 with your User-Agent, 403 with a real Chrome UA,
403 with none, and old.reddit.com 302s away. Every test here stubs fetch,
so the suite passes while the integration returns an HTML error page.

plugins/reddit/search.js already searches Reddit using Strategy.COOKIE with
browser: true — that's the pattern for Reddit in this repo, because the
server-side JSON path is blocked.

Also, adding reddit to the *default* source list means perPlatform =
ceil(limit / 7) instead of /6, so the six working sources each return fewer
results while Reddit contributes nothing. allSettled keeps it from
crashing, but `research --limit 20` gets worse.

Happy to merge a PR with just the npm changes plus the two fixes above.

PR #426, opened 2026-08-23
fix(observation): redact Basic/Negotiate/NTLM auth credentials in raw text, not just Bearer

arzaanxeng:
## What

`redactText()` in `src/observation/redaction.ts` only stripped the `Bearer`
auth scheme before writing raw text into observation artifacts. Basic,
Negotiate, and NTLM credentials passed through untouched:

    redactText('Authorization: Basic dXNlcjpwYXNzd29yZA==')
    // -> 'Authorization: Basic dXNlcjpwYXNzd29yZA==' (unredacted)

    redactText('Authorization: Bearer abc.def.ghi')
    // -> 'Authorization: Bearer [REDACTED]' (correctly redacted)

## Why it matters

`redactText` is the last line of defense for raw text that ends up in
stored observation artifacts — extracted page content, page titles, and
rendered command output (see `browser/run/runner.ts` and
`browser/runtime/local-cloak/actions.ts`) — as opposed to structured
header objects, which already go through `redactHeaders`. A page
displaying an HTTP Basic Auth header (API docs, a debug/status page, a
network trace echoed into visible content) would have that credential
captured verbatim in stored artifacts.

## Fix

Generalized the Bearer-only regex to cover the standard RFC 7235 auth
schemes (Bearer, Basic, Negotiate, NTLM), preserving the matched scheme
name in the redacted output:

    .replace(/\b(Bearer|Basic|Negotiate|NTLM)\s+[A-Za-z0-9\-._~+/]+=*/gi,
      (_match, scheme) => `${scheme} [REDACTED]`)

## Testing

- Added a regression test covering Basic, Bearer, and NTLM in
  `redaction.test.ts`.
- Ran the full unit suite before and after: 154 files / 2627 tests,
  all green, no regressions.

### DIFF
--- a/src/observation/redaction.test.ts
+++ b/src/observation/redaction.test.ts
@@ -1,7 +1,18 @@
 import { describe, expect, it } from 'vitest';
-import { redactHeaders, redactUrl, redactValue } from './redaction.js';
+import { redactHeaders, redactText, redactUrl, redactValue } from './redaction.js';
 
 describe('observation redaction', () => {
+    it('redacts Basic/Negotiate/NTLM auth credentials in raw text, not just Bearer', () => {
+      expect(redactText('Authorization: Basic dXNlcjpwYXNzd29yZA=='))
+        .toBe('Authorization: Basic [REDACTED]');
+      expect(redactText('curl -H "Authorization: Basic dXNlcjpwYXNzd29yZA==" https://api.example.com'))
+        .toBe('curl -H "Authorization: Basic [REDACTED]" https://api.example.com');
+      expect(redactText('Authorization: Bearer abc.def.ghi'))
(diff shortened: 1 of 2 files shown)


arzaanxeng:
This is an internal fix to the redaction regex (no public API or user-facing behavior change), so I don't think it needs a docs/skill update — happy to add one if you disagree. Also flagging that the automated review returned an invalid result rather than a verdict, so this may need docs-not-needed applied manually.

adikulkarni006 (approved):

ankitranjan7:
Thanks — confirmed against main: redactText was still Bearer-only, and the Basic/NTLM/Negotiate cases fall through into saved observation artifacts. Merging.

arzaanxeng:
Thanks for the review and for confirming it against main!

PR #461, opened 2026-08-28
fixed title and sidebar title names

rajarshidattapy:


rajarshidattapy:
Added docs for cline and aider.

